PR #21, opened 2016-02-09
Trello tasks 0402

mkroon1:
This PR fixes issue on Trello card: "A curator's VE has nice list of genes that expands. Let's build the same for the shared column's VE (diseases or genes). Use code from curator's VE."

It also concerns setting mysql concat_group_max_len setting in a new query run before viewEntry or loadEntry.

Note before merging: links generated for the custom columns point to the genes/diseases for now. They can be changed to point to the custom column data for that specific gene/disease. Let me know if that's needed.

### DIFF
--- a/src/class/object_columns.php
+++ b/src/class/object_columns.php
@@ -5,7 +5,7 @@
  *
  * Created     : 2010-03-04
  * Modified    : 2016-02-08
- * For LOVD    : 3.0-14
+ * For LOVD    : 3.0-15
  *
  * Copyright   : 2004-2015 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ing. Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
--- a/src/class/object_columns.php
+++ b/src/class/object_columns.php
@@ -77,15 +77,15 @@ class LOVD_Column extends LOVD_Object {
                                            '(a.colid IS NOT NULL) AS active, ' .
                                            'uc.name AS created_by_, ' .
                                            'ue.name AS edited_by_, ' .
-                                           'GROUP_CONCAT(sc.geneid ORDER BY sc.geneid SEPARATOR ";") AS _geneids,' .
-                                           'GROUP_CONCAT(sc.diseaseid ORDER BY sc.diseaseid SEPARATOR ";") as _diseaseids';
+                                           'GROUP_CONCAT(sc.geneid ORDER BY sc.geneid SEPARATOR ";") AS _genes, ' .
+                                           'GROUP_CONCAT(DISTINCT d.id, ";", IF(CASE d.symbol WHEN "-" THEN "" ELSE d.symbol END = "", d.name, d.symbol) ORDER BY (d.symbol != "" AND d.symbol != "-") DESC, d.symbol, d.name SEPARATOR ";;") AS __diseases';
         $this->aSQLViewEntry['FROM']     = TABLE_COLS . ' AS c ' .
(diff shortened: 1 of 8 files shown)


src/class/object_columns.php
```diff
@@ -66,17 +66,26 @@ function __construct ()
...
+
         // SQL code for viewing an entry.
         $this->aSQLViewEntry['SELECT']   = 'c.*, ' .
                                            'SUBSTRING_INDEX(c.id, "/", 1) AS category, ' .
                                            'SUBSTRING(c.id, LOCATE("/", c.id)+1) AS colid, ' .
                                            '(a.colid IS NOT NULL) AS active, ' .
                                            'uc.name AS created_by_, ' .
-                                           'ue.name AS edited_by_';
+                                           'ue.name AS edited_by_, ' .
+                                           'GROUP_CONCAT(sc.geneid) as geneids,' .
```
ifokkema: You can make LOVD work a bit for you, if you use code like in object_users.php:
`GROUP_CONCAT(sc.geneid ORDER BY sc.geneid SEPARATOR ";") AS _geneids`
Then, you don't have to worry about exploding it yourself, LOVD will do it for you, because of the combination of the underscore in front of the column name, and the semicolon as separator. LOVD will then create $zData['geneids'] for you, already exploded. Also, the ORDER BY may be important. Could you please make these changes for both GROUP_CONCATS?

src/class/object_columns.php
```diff
@@ -397,6 +406,24 @@ function prepareData ($zData = '', $sView = 'list')
...
             $zData['public_add_']      = '<IMG src="gfx/mark_' . $zData['public_add'] . '.png" alt="" width="11" height="11">';
             $zData['allow_count_all_'] = '<IMG src="gfx/mark_' . $zData['allow_count_all'] . '.png" alt="" width="11" height="11">';
+
+
+            if ($zData['category'] == 'VariantOnTranscript') {
+                // Show genes for which this column is activated.
+                $this->aColumnsViewEntry['related_genes'] = 'Column activated for genes';
+
+                $aGenes = explode(',', $zData['geneids']);
+                $zData['related_genes'] = LOVD_Object::lovd_getObjectLinksHTML($aGenes, 'genes/%s');
```
ifokkema: The advantages of using $this->lovd_getObjectLinksHTML() are slightly higher, please update the LOVD_Object::lovd_getObjectLinksHTML() calls to using $this.

src/class/objects.php
```diff
@@ -1518,5 +1530,31 @@ function viewList ($sViewListID = false, $aColsToSkip = array(), $bNoHistory = f
 
         return $nTotal;
     }
+
+    public static function lovd_getObjectLinksHTML($aIDs, $urlFormat)
+    {
```
ifokkema: Oops, I forgot to tell you... functions and methods are sorted alphabetically... could you sort this function to its correct location in the file? Also, methods are separated by 5 new lines.

src/class/objects.php
```diff
@@ -1518,5 +1530,31 @@ function viewList ($sViewListID = false, $aColsToSkip = array(), $bNoHistory = f
...
+        // Returns a list of gene links in HTML.
+        // Parameter $genes is an array with gene symbols as values.
+        // Parameter $urlPattern designates the target of the links, where %s
+        // will be substituted with the gene symbol.
+
```
ifokkema: Please update the function's description; $genes has become $aIDs, and function is not only for genes now.
Please also update $urlPattern to $sURLPattern or $sUrlPattern, whichever you prefer :)

ifokkema:
Excellent, thanks!
I've made some comments attached to certain lines. In general, could you please also update the copyright statements? Thanks!

mkroon1:
@ifokkema, I attended to the comments above, it should be ready to merge now.

PR #28, opened 2016-02-16
Bug/link uncurated variants

### LINKED ISSUE #26: Wrong link for uncurated variants in drop-down menu Configuration area
Copied from Trello:

> Data entries can have different statuses (see inc-init.php for the list, STATUS_* constants). See the manual (search for "statusid") for their explanations.
Marked is a bit special, because it's sometimes considered both Uncurated and Curated. It's public, but it's marked because the curator needs to do something with it still. In counts, Marked is considered curated (public), because otherwise the numbers would not add up.
The Configuration area has a variant information block on the left with numbers, and links. See the links and check how they work, by injecting search terms into the ViewList by setting a $_GET argument.
Now, the Configuration area tab's dropdown menu, has the option "View uncurated variants". "Uncurated" in this sense, means "Pending" (previously called "Submitted"), "Non public", or "Marked". As you may notice, the link in the dropdown menu doesn't work properly. It doesn't show all uncurated variants.

mkroon1:
Note that in re-wording the link text for uncurated variants, I also moved the link down one step in the drop-down list, as I think it's prettier now that the text of the two links is very similar. This PR resolves #26

### DIFF
--- a/src/class/template.php
+++ b/src/class/template.php
@@ -4,13 +4,14 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2012-03-27
- * Modified    : 2016-02-02
+ * Modified    : 2016-02-15
  * For LOVD    : 3.0-15
  *
  * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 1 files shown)


ifokkema:
Excellent, thanks!

PR #30, opened 2016-02-17
Prevent browsers from auto-filling usernames in data entry forms

### LINKED ISSUE #27: Issues with browsers remembering user's username and password
From Trello:

> Some people like to configure their browsers to remember their password when logging into LOVD. We can't stop them, and I guess we shouldn't, but this has annoying side effects that we should try to stop.
The browser is not the smartest thing and thinks that every form that has a password field, is a login form. As such, it injects the password in the password field (alright), but also tries to find a field to put the username in. The browser is stupid and injects the username in the last field that is used before the password field. This can be any field in LOVD, and as such we see user's usernames showing up in submissions all the time.
Devise a plan that would be simple to implement in LOVD, that would fix this problem. Examples: having LOVD add a hidden text field that does nothing (field's name is important) right before a password field, or perhaps field with display:none, or otherwise detect user's username in last field and automatically remove?

mkroon1:
Improvements to getForm() form generator in inc-lib-form.php. Blocking the autofill browser feature for password fields is off by default. All password fields are preceded by a hidden text field to avoid mistakenly autofilling the username in some unrelated text field.
Resolves #27

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -59,11 +59,15 @@
    is active for, is now shown on the column's detailed view.
  * Fixed bug; Transcript viewlists included non-public variants in the variant
    counts also for non-authorized users.
- * Fixed bug; automatic filling of certain password fields in forms is now
+ * Fixed bug; Automatic filling of certain password fields in forms is now
    blocked and a fake text field is used to capture auto-filled usernames.
-   Closes #30
- * Fixed bug; fixed link in drop-down under Configuration tab to variants that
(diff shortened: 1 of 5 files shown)


ifokkema:
Excellent, thanks! Could you please add a changelog entry?
We also miss a changelog entry for #28, I forgot to remind you :)

mkroon1:
@ifokkema Changelog is updated.

ifokkema:
Thanks!

Issue #7, opened 2016-01-20
Check referring numeric IDs in import files


dasscheman:
The {{id}} field in the Individuals section needs to be numeric, but the {{individualid}} in the Individuals_To_Diseases and Screenings sections is not checked this way. Also, it doesn't seem to complain if the individualid given is found nowhere (not in the file, not in the database).

PR #31, opened 2016-02-18
Import

### LINKED ISSUE #6: Mesage: Please select a valid entry from the 'refseq_genomic' selection box should not be displayed.
In object_genes the variable $aSelectRefseqGenomic is not set properly. It is only filled with the first NC that is found, hereafter the caching in the getForm function  (if (!empty($this->aFormData)) { return parent::getForm(); }) cause that this variable is not further completed.

dasscheman:
Trello tasks https://trello.com/c/G87b1p0c and https://trello.com/c/x5HLAegf
See commits for explanation. 

**edit by mkroon1:**
Fixes #6 (but not #7)

### DIFF
--- a/src/class/object_genes.php
+++ b/src/class/object_genes.php
@@ -296,10 +296,7 @@ class LOVD_Gene extends LOVD_Object {
                 global $zData;
                 $aSelectRefseqGenomic = array_combine(array($zData['refseq_genomic']), array($zData['refseq_genomic']));
          
-                $this->aFormData =
-                    array(
-                        array('Genomic reference sequence', '', 'select', 'refseq_genomic', 1, $aSelectRefseqGenomic, false, false, false)
-                    );
+                $this->aFormData['refseq_genomic'] = array('Genomic reference sequence', '', 'select', 'refseq_genomic', 1, $aSelectRefseqGenomic, false, false, false);
             }
             return parent::getForm();
         }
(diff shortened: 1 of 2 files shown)


ifokkema:
Hello,
I made some comments on the commits, could you please attend to those, if you have time?
The second trello card (https://trello.com/c/x5HLAegf) doesn't actually seem to be included? Although one of the issues doesn't seem to happen anymore (not complaining if the individual isn't found), that also doesn't happen in master anymore, and the individualids are still not forced to integers. Try this import file, so you see what I mean:
```
### LOVD-version 3000-14c ### Full data download ### To import, do not remove or alter this header ###
# charset = UTF-8

## Diseases ## Do not remove or alter this header ##
## Individuals ## Do not remove or alter this header ##
"{{id}}"
"asdf"

## Individuals_To_Diseases ## Do not remove or alter this header ##
## Count = 16
"{{individualid}}"	"{{diseaseid}}"
"asdf"	"00001"
```
The "id" field is checked properly and complains that "asdf" is not a number, but LOVD does not complain about "individualid" is not numeric.

dasscheman:
Ah now I see what you mean. I thought it was about the first commit; integers checked as strings. 
I will have a look.

dasscheman:
It seems to be a more general problem, because for transcriptid, screeningid, variantid this also happens. Changing diseaseid to 'asdf' gives an mysql_error. I'm not sure if you want this checked in objects, or in the import.

ifokkema:
OK, I'll take over and take a look at it later, then :)

PR #32, opened 2016-02-18
Improved logging for management of curators for genes.

### LINKED ISSUE #25: Improve logging when changing curator rights
When changing curators or curator's rights (genes/DMD?authorize), LOVD logs that the form is submitted, but not what has changed. It would be useful to change the lovd_writeLog() call to include the difference, so you can see in the logs if a user was removed as curator, demoted to collaborator, collaborators who became curators, or users added as curators.

- [x] First make a plan on how this log entry would look for all possible situations, so we can discuss
- [x] Then implement by creating something like $sAuthorizeDiff and including that in the lovd_writeLog() function call

mkroon1:
Now logs:
 * Changes to edit privileges
 * Shown/hidden curators
 * Order of curators

Resolves #25

### DIFF
--- a/src/class/object_logs.php
+++ b/src/class/object_logs.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-01-28
- * Modified    : 2016-02-18
+ * Modified    : 2016-02-22
  * For LOVD    : 3.0-15
  *
  * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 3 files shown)


ifokkema:
Thanks! I've added some comments. Could you please attend to them, and add a changelog entry? Thanks!

mkroon1:
@ifokkema, thanks for the comments, I've attended all. Please note:
 * I've introduced a new constant LOVD_EOL as a project-wide default for end-of-line characters. I thought this would be handy as I already saw "\n" scattered here and there.
 * I managed to refactor the code using 1 query, but it's become quite clumsy -- feel free to improve it.

ifokkema:
Thanks!!! I've been nitpicking through your code a bit (sorry!) and added comments here and there :wink:

mkroon1:
> Thanks!!! I've been nitpicking through your code a bit (sorry!)

That's OK, I'm breaking things here and there anyway. And it's nice for you to thank me, but you don't have to thank me for every separate commit :wink:

ifokkema:
Alright, I just did some minor changes. Let's fix the issue with $aCurators possibly being empty, and then we can merge this PR :smile:

PR #33, opened 2016-02-18
Added double warning before deleting gene or disease.

### LINKED ISSUE #29: No proper warnings when deleting a gene or a disease
From Trello:
> When deleting a gene, all VOT (VariantOnTranscript) entries are removed. When deleting a disease, all phenotype entries are removed. The latter is not mentioned when deleting a disease. And both forms just ask for the password once, while LOVD2 used to ask for double confirmation.

- [x] Download the LOVD2 code, and check setup_genes.php. Look for the code that handles the $_GET['drop'] variable, and you'll find the code that handles deleting a gene. Note how a second variable, $_GET['confirm'], is used to make the user think twice about deleting the gene.
- [x] LOVD3 is slightly different, variants are not removed, but their mapping on the gene is removed. This warning is already present. Now, implement the double form method in LOVD3 as well, and display on that page the number of variants that will lose their mapping to this gene.
- [x] Add to the page that deletes a disease a similar warning that the gene removal form already has, about all phenotypes being removed.
- [x] Implement the double form method for deleting a disease, too.

mkroon1:
@ifokkema  Please check whether warning messages are clear and informative.

Resolves #29

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_script:
   - ./tests/travis/setup/setup_apache.sh --githubaccount=LOVDnl
   ## IMPORTANT! convert must be run AFTER the setup_apache.sh
   ## setup_apache.sh fix some assumptions where the files are stored!
-  - ./tests/bash_scripts/convert_selenium_to_phpunit.sh --projectfolder=LOVD3 --continueall
+  - ./tests/bash_scripts/convert_selenium_to_phpunit.sh --localhost=http://localhost/LOVD3 --continueall
   - ./tests/travis/setup/setup_config.sh
   - ./tests/travis/setup/setup_selenium_server.sh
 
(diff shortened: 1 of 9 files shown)


ifokkema:
I like how you made the changes without adding much new code! The additional nesting as done in LOVD2 seems like quite some overkill. However, now it lets me go to the second form with no password filled in, or the wrong password filled in.
Also, what I liked most about the LOVD2 method was the statistics that were shown, so you would know what you would delete with the gene or disease.
Could you please fix the issue with the password not being checked, add the statistics (basically mention all that will be lost; transcripts and VOTs in the case of genes, and phenotype entries in the case of diseases), and add a changelog entry? Once done, we need to also update the tests, because we just broke them ;)
Thanks!

mkroon1:
@ifokkema, I've fixed the password authorization issue and added some stats. Please let me know if you agree with the text before I fix the test.

ifokkema:
Hi! I had some issues with running it, and I thought it would be easier if I made some changes myself. Please check my commit and my commit message to see if it's all clear or not. If you agree to the changes, could you then please fix the tests? If that's done, feel free to merge the branch to master :relaxed:

PR #34, opened 2016-02-23
Added a git pre-commit hook for checking *.php file headers.

mkroon1:
pre-commit is run just before a commit is made. The script checks all added/modified php files for their headers. Currently it checks for the existence of the modification date (current date) copyright year (current year) and author (name/email from the git config). If one of the checks fails the commit is aborted.

The pre-commit hook requires python 3.  Install the hook by running create_hook_symlinks.sh which will create a symlink in .git/hooks

@ifokkema, please see if it's useful for you, currently it checks for your name/email in the file header based on the git config -- which is different from what's in the file header in your case I believe. If you're not going to use it, we may as well keep it out of source control.

### DIFF
--- a/git_hooks/create_hook_symlinks.sh
+++ b/git_hooks/create_hook_symlinks.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Create symlinks in .git/hooks for hook scripts in this directory.
+# The script assumes it is executed in the directory where the hook
+# scripts are that need to be linked.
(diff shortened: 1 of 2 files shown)


ifokkema:
Nice! I can imagine this is useful for others as well. If it doesn't match my settings, I can either modify my settings or choose to not use the hook ;)
I just had one mini comment. Feel free to merge this PR with master!

PR #40, opened 2016-03-11
Removed css class deletion for form field VariantOnGenome/DNA

### LINKED ISSUE #39: Transcript variant entry form: warning on empty *Genomic DNA change* unclear
In the variant entry form (url ends with `variants/ID?edit`), the field *Genomic DNA change* is required. But when left empty, it's not highlighted after the form is submitted (though a warning is shown at the top of the page).

Additionally, such errors may be avoided in some cases when the content for *Genomic DNA change* is automatically generated from the field *DNA change* without having to manually click *Map to genome*. Is such behavior always appreciated, or does one in some cases override the value for *Genomic DNA change*?

mkroon1:
All css classes were being deleted in javascript for input field VariantOnGenome/DNA. This broke the error reporting. I removed the deletion, as it didn't seem to have purpose (the function doesn't set a class attribute for the field anywhere else).
Closes #39

### DIFF
--- a/src/inc-js-variants.php
+++ b/src/inc-js-variants.php
@@ -4,14 +4,14 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2011-11-08
- * Modified    : 2016-02-05
+ * Modified    : 2016-03-14
  * For LOVD    : 3.0-15
  *
  * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ing. Ivar C. Lugtenburg <I.C.Lugtenburg@LUMC.nl>
  *               Ing. Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
  *               Msc. Daan Asscheman <D.Asscheman@LUMC.nl>
- *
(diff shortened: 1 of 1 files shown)


ifokkema:
Ah... the purpose is to remove any error classes, that may have been put there while manually mapping the variant using the form. I guess it's missing some documentation there. Check for classes 'warn' and 'err' in inc-js-variants.php, and you'll see it being set. I guess we need to find a way to make sure we only remove those classes when needed. When the form is freshly loaded, none of the JS functions should run, right? Any idea why lovd_checkHGVS() is triggered?

mkroon1:
That makes sense.

> Any idea why lovd_checkHGVS() is triggered?

It's triggered when te page is loaded. Probably to check the fields when an existing variant is being edited. I've changed it now to only call lovd_checkHGVS() on page load when the field is not empty. After that, when the field is edited, the 'err' class will again be removed, but that's expected behavior in my opinion.

ifokkema:
Ah, right, I remember now. We run that function to enable the mapping buttons. Your suggestion makes sense, thanks! I do see minor issues in the code, I will add some comments there. When they are addressed, feel free to merge this into master!

PR #46, opened 2016-03-21
Refactor tests

mkroon1:
A refactoring of the selenium tests, which includes the following:
* Deleted all Selenium IDE test files.
* Removed conversion script (from selenium to phpunit).
* Have all tests derive from LOVDSeleniumBaseTestCase.
* Introduced phpunit.xml for configuring test suites.
* Have Travis run test suites directly with phpunit.
* Let Travis upload screenshots if build fails.

Caveats:
* Tests within a test suite are usually not independent, they cannot be run separately.
* Some tests are shared between test suites, these are located in `shared_tests/`.
* One cannot use phpunit directly to run all tests, as phpunit fails when it tries to load certain classes multiple times. You can use `run_all_tests.sh` for this purpose.


ifokkema:
Merged! Sorry to have been slow, somewhere in my head I thought this wasn't fixed/done yet...

PR #49, opened 2016-03-23
Feature viewlist find replace

### LINKED ISSUE #36: Find & Replace in viewlists
From Trello:
> LOVD2 had a Find and Replace method. It allowed, from a VL or directly by opening a form, to apply filters on your gene's data set, and have a certain value in a column replaced by another. However, you could never see the effect of your actions until they were done. In LOVD3, we must do this better, and we'll show the effects of the Find & Replace before it's run.
A good idea is to keep the code for F&R in MySQL, so you can create a VL and all, with differences highlighted. MySQL can select both the original column and the modified column, and thus LOVD can show which values have changed.

- [x] Find & Replace should only be available for selected VLs
- [x] Find & Replace is a new option in the options menu
- [x] After selecting, it will write a message to the screen (dialog or otherwise) that tells the user to select the right column. On clicking on the column, the value is then stored in the VL form and the VL is submitted
- [x] F&R can only be run on custom columns, not on any other columns. When selecting your column to run F&R on, this should be visibly marked.
- [x] Styles.css should be edited to allow the replaced column to be marked.
- [x] New VL carries the form from LOVD_Object::viewList(), with selection list with options (replace anything (replace entire column's value), match anywhere in field, match at beginning of field, match at end of field), text to replace, text to replace with. Form has button to show changes, and to apply changes.
- [x] LOVD_Object::VL() should alter query to add additional column with updated value. New function should be written to generate REPLACE() or regular expression in SQL to perform the F&R.
- [x] When "Show changes" is pressed, the VL is submitted. When the "Apply changes" is pressed, a dialog opens with a confirmation form with the number of changes reported, and asking for the user's password for confirmation. The "Apply changes" button should be red or otherwise clearly marked as "dangerous".
- [x] Required level is LEVEL_CURATOR.

## Code review:
- [x] Add "Cancel" option, when clicking F&R option. Now you must select a column to be able to cancel. Could be triggered by clicking outside of table, pressing `Esc` or perhaps a button.
- [X] Prepare for allowing more than just custom columns (we'll deal with actually allowing them later)
- [x] Enable F&R also for other VLs, not just Screenings.
- [x] ~~Don't include JS for F&R when not authorized.~~
- [x] Check all `lovd_displayError()` calls and make sure they have proper Error names defined (no long constant-like values)
- [x] Processing a F&R should not be done through GET, because it is a data update (CSRF!!!), and it should contain a password which should not be put in a query string. Better separate it completely from the VL. It can be an Ajax script that processes the F&R, then sends back success yes/no or any other info needed, after which the VL can be reloaded by the interface.
- [x] I managed to lose the preview button at a certain point; is there a logic that hides it sometimes?
- [x] When you cancel a preview, the preview column stays. You can select this column for a new F&R, preview that, and cause a query error.
- [x] The use of the field name with or without the table name, and the use of the table name, is inconsistent. This causes flaky performance (i.e. random failures) when trying to let non-custom columns work (or when refactoring the code, like for me :wink:).
- [x] Nice to have: Functions shouldn't really return void when they error. If something is wrong, they should return false. If the function should return some kind of data structure, we need to check how to implement this properly.
- [x] The edited_date and edited_by fields should be edited, as well.

mkroon1:
Column-wise find & replace for viewlists.
Closes #36

### DIFF
--- a/git_hooks/create_hook_symlinks.sh
+++ b/git_hooks/create_hook_symlinks.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Create symlinks in .git/hooks for hook scripts in this directory.
+# The script assumes it is executed in the directory where the hook
+# scripts are that need to be linked.
+# Code is inspired by: 
+# http://stackoverflow.com/questions/3462955/putting-git-hooks-into-repository
+
+HOOK_NAMES="applypatch-msg pre-applypatch post-applypatch pre-commit prepare-commit-msg commit-msg post-commit pre-rebase post-checkout post-merge pre-receive update post-receive post-update pre-auto-gc"
+
+# Location where git expects the hooks.
+HOOK_DIR=$(git rev-parse --show-toplevel)/.git/hooks
+
+# Script location (it's best to make this relative from the HOOK_DIR)
+SCRIPT_DIR=../../git_hooks
+
+for hook in $HOOK_NAMES; do
+
+    if [ -a $HOOK_DIR/$hook ]; then
+        echo "Hook $hook already exists, skipping!"
+        continue
+    fi
+
+    if [ -x $HOOK_DIR/$SCRIPT_DIR/$hook ]; then
+        # Executable hook script found. Creating a symlink for it.
+        cmd="ln -s $SCRIPT_DIR/$hook $HOOK_DIR/$hook"
+        echo $cmd
+        $cmd
+    fi
+done
--- a/git_hooks/create_hook_symlinks.sh
+++ b/git_hooks/create_hook_symlinks.sh
@@ -16,12 +16,11 @@ SCRIPT_DIR=../../git_hooks
 
 for hook in $HOOK_NAMES; do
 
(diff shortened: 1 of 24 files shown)


ifokkema:
Hi Mark, I'm not sure how this happened, but there's lots of changes in this PR that are not related to Find & Replace. It's really difficult for me to look at the code/PR like this. Could you see if you can find out what happened, and see if we can remedy this situation somehow?

mkroon1:
The mess is due to (mis-)use of git merge and rebase. This branch was diverging from master, so I merged changes from master to here. Initially this was presented as a single commit (at least in github's display). But later I did a rebase in this branch on itself, in order to clean up some early commits by squashing them. This resulted in the merge to be displayed as all separate commits.

The commits are not completely in disorder though. As far as I can see the find & replace commits are together in 2 groups: 48ca406 through 9447b2b and f486369 through d5b2377 and a single commit in the middle: 899c3e9.

I guess the easiest way to review is to do diffs on these ranges. If you want I can do another rebase and squash all non-F&R commits to create some clarity.

PR #68, opened 2016-05-17
Feat/share permissions

### LINKED ISSUE #51: Sharing permissions
## Feature

A new feature to allow submitters to share the view/edit permissions on the
objects that they own with other users.

For two users, A and B.

* A, as owner or creator of X, wants to share his (viewing or editing) permissions with user B, so that B can help maintain X. Where X is any object created by or owned by A of the following types: individuals, phenotypes, screenings, variants (all objects that have an owned_by field).
* A shares his viewing or editing permissions with *colleague* B. A decides whether or not B is allowed to receive edit permissions, using a checkbox on the form.
* B can now view or edit all objects owned by A.
* B can not view or edit non-public objects *created by* A, unless they are also *owned by* A.
* B receives an email stating that A shares permissions with him for all A's objects and instructions how to share permissions with someone.

## Implementation outline

- [x]  New table `TABLE_COLLEAGUES` (user_id1, user_id2), where user_id1 shares his permissions with user_id2 (column names `userid_from`, `userid_to`?).
- [x] Shared permissions defined in `TABLE_COLLEAGUES` are 1-directional.
- [x] New table should also be created on upgrade.
- [x] Do we need to do something about the fact that this update will cause query errors on the data views until the upgrade process has been run?
- [x] Then, do some speed tests on real data to see if sending a large list of colleagues for a VL will slow down the VL a lot. Try both sending IDs through a JOIN, as sending IDs through as arguments. See how LOVD currently does this for non-public entries and submitters being logged in.
  - [ ] If that is too slow, we could either not modify the VLs, maybe only modify the gene-specific VLs, or let the user choose to turn this setting on, warning the user of a slow down? Discuss this with Johan.
  - [x] If it doesn't slow down the VL a lot, we'll implement it in the VLs as well.
- [x] View entry of user page (`users/user_ID`, reached by clicking "Your account") shows users with which authentication is shared.
- [x] Options menu on user page has link to page for sharing permissions.
- [x] Page for sharing permissions is similar to the curator management page on `genes/geneID?authorize`.
- [ ] Submitters should be able to see the users overview? It's kind of weird that they can read all users when giving users rights, but not in general just get a list? Check if fields displayed are OK.
- [x] Users who share their permissions with the current user are retrieved when loading $_AUTH.
- [x] Adapt all queries where now the data status is checked and the ownership is checked, to also include a check for colleague level, using the owned_by field only (not created_by).
- [x] lovd_isAuthorized() is extended to take shared permissions into account (it will change the user level to LEVEL_COLLEAGUE based on shared permissions).
  - [x] Make sure it uses owned_by only, and not created_by, and that it checks for and uses the allow_edit flag.
- [x] Review all data view entries to check their requireAuth() calls and see if they're still OK.
- [x] A user who is a recipient of shared permissions will get an e-mail noting so and instructions on how to share permissions themselves (optionally with a direct link to share permissions back to the original user who shared their permissions with them. -- I.e. "returning the favor").

## Code review:
- [x] Make the curator selection and selecting colleagues match better (using different JS functions now).
- [x] Fix bug that VLs no longer search in chromium < 49 and some other browsers.
- [x] Fix bug that colleagues can not be put back in the VL (VL should remove the IDs from the search box).

mkroon1:
Fixes #51


mkroon1:
Looks to me we're ready to merge

Issue #69, opened 2016-05-17
Clarify IP allow list functionality on form and confirmation email


ifokkema:
Spend a little bit of time improving the submitter registration form by warning the user a bit more elaborate that the `Allowed IP address list` can block them out of their account.
- [x] Add note to form, either in JS or otherwise, that warns the user a bit more, if something has been filled in that is not `*`.
- [x] Mark the field as optional.
- [x] Add field's contents to registration confirmation email.
- [x] Add part in the confirmation email that warns the user, if this field has been filled in.
- [x] If the user is blocked based on IP address, provide information of managers (like in Configuration area) to allow to ask for help.
- [x] Optionally, see if we can make sure LOVD does not suggest IP6 addresses, since we don't support them.

PR #81, opened 2016-06-09
Clarify IP allow list feature

ifokkema:
The IP allow list feature needed clarification.
- Clarified that the IP address list is optional.
- Clarified that the IP address list field's default value is *.
- Added JS notice with a warning when user sets the IP address allow list.
- When the form warns about the field being filled in, and you press "Cancel", the IP address list field receives focus.
- Added the IP addresses to the email sent to the user after registering.
- Added a message to the email body when the IP address list has been set to something other than '*'.
- When user is blocked by IP, the manager info is shown.  …
- This, so that the user can know who to contact to get help.
- Clarified we don't support IPv6 for restricting access:  …
  * The form won't show your current address any more, when it's IPv6.
  * When it's IPv6, no input in the IP address list field will be allowed, and LOVD will tell you it's because we don't support IPv6.

Implements #69.

### DIFF
--- a/src/class/object_users.php
+++ b/src/class/object_users.php
@@ -274,7 +274,14 @@ class LOVD_User extends LOVD_Object {
                 // Check given security IP range.
                 if ($bIP && !lovd_validateIP($aData['allowed_ip'], $_SERVER['REMOTE_ADDR'])) {
                     // This IP range is not allowing the current IP to connect. This ain't right.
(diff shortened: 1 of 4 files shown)


ifokkema:
@mkroon1 Wil jij nog naar de code kijken voordat ie gemerged wordt? De failures zijn omdat Mutalyzer een fout heeft in een webservice.

PR #92, opened 2016-07-05
Added remarks field for transcripts.

### LINKED ISSUE #77: Comment on transcript
Dear Ivo,

I wanted to see if it is easy to add a comment section on the transcript page . As a transcript selected for a gene may not always be the longest one and may be selected on a different criteria, it would be advantageous to include some information on why the transcript was selected. At the moment I can't find a way to include this information on the transcript section( but possible on variant on transcript)

Warm regards,
Rahul

## Tasks
- [x] Adapt inc-sql-tables.php to include new column.
- [x] Adapt inc-upgrade.php to include new column.
- [x] Describe field in getForm() in object_transcripts.php (as an optional textarea, like other remarks fields).
- [x] Describe field in VE column list in object_transcripts.php.

mkroon1:
Fixes #77

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -11,6 +11,9 @@
  * Fixed bug; Impossible for users of level submitter to browse the user list
    on the access sharing page.
  * Fixed bug; Repeated gene names on user acccount pages.
+ * Added a "remarks" field for transcript. This may for example be used to
+   denote why a certain transcript was chosen.
+   Closes #77: "Comment on transcript".
 
 
 /**************************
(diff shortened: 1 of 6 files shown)


mkroon1:
I changed my mind regarding how the accepted form fields for editing transcripts are defined. It's not within the scope of this pull request to get the fields from `getForm()` for every call to `updateEntry()` for all objects.

Still, in the future I think it would be nice to have something like `buildFields()` for non-custom objects.

ifokkema:
Alright, all good to merge then!
(although there are conflicts)

PR #93, opened 2016-07-12
Feat/selenium webdriver

mkroon1:
Upgrade to selenium webdriver


ifokkema:
Readme looks good! Not having tested it though, I wonder how composer installs phpunit? The instructions say to install Selenium Server and FF, how to start Selenium, and then to execute PHPUnit. Should more info be included on how to get composer to install all needed packages? You'll understand from this that I never used composer :grin:

mkroon1:
Composer is not essential for running the tests, so I only mentioned it briefly for now.

If you run `composer install` in the main project directory, it will install the dependencies to a `vendor/` subdir where you can find the phpunit binary. For more information I'd like to refer you to the composer website: https://getcomposer.org/ :smile:

PR #103, opened 2016-07-21
Fix/owned by len

### LINKED ISSUE #65: Restrict length of owned_by selection field
When the "owned_by" selection field gets too long because of long user's names, the form layout is distorted by the browser. This does not look very good.
Somehow make sure that the field has a maximum width, solving this problem.

mkroon1:
* For forms generated by lovd_viewForm(), the holding table has a 'fixed'
  lay-out, so that it doesn't get distorted by its content.
* Select input fields have max-length set to 100%, making sure they won't
  overflow parent elements.
* Adjusted column widths for the major edit/create forms (tested on 
  chromium + firefox)

Fixes #65


mkroon1:
@ifokkema, I remember there was talk about making sure `lovd_shortenString()` was called with at most 75 characters for diseases in edit forms. I didn't see any instances where it is wrong -- do you know if this is still an issue?

ifokkema:
In `columns.php`, `lovd_shortenString()` is used in 4 places, when building the forms for adding and removing custom columns. There, the function is called without providing a length argument. The default is currently 50 characters, only used in forms for the genes. Diseases shorten at 75. What we could do is shorten the genes as well at 75 (updating the forms to match this change), and perhaps just change the default value? Since we never call `lovd_shortenString()` without a second argument anywhere else but there. Otherwise, change the code in `columns.php` to shorten at 75 for diseases? Or just keep the inconsistency...

mkroon1:
> Otherwise, change the code in columns.php to shorten at 75 for diseases?

I opted for this solution.

PR #104, opened 2016-07-22
Let import check the import file first for panels (of individuals).

### LINKED ISSUE #67: Import: Panel_size of individual in panelid field not checked properly
When importing multiple individuals where the `panelid` field of one of them refers to an individual in the file, LOVD does not properly check the `panel_size` field of the individual referred to.
File example:
```
### LOVD-version 3000-150 ### Full data download ### To import, do not remove or alter this header ###
## Individuals ## Do not remove or alter this header ##
{{id}}	{{panelid}}	{{panel_size}}
1		4
2		1
3	1	1
```
Returns error message:
```
Error (Individuals, line 6): Panel ID "1" refers to an individual, not a panel (group of individuals). If you want to configure that individual as a panel, set its 'Panel size' field to a value higher than 1.
```
Which is not correct.

mkroon1:
Fixes #67 

@ifokkema: Let me know if this solves #67 to your satisfaction, see the discussion in the issue's comments.

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -16,6 +16,9 @@
    Closes #77: "Comment on transcript".
  * Fixed bug; Query errors could occur when performing incorrect searches in
    hidden columns.
+ * Fixed bug; Error when updating individuals panel from import file.
+   Closes #67: "Import: Panel_size of individual in panelid field not checked
+   properly"
 
 
 /**************************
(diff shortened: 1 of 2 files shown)


ifokkema:
Cool, thanks! This means that from now on, if you want to link a new individual to an existing panel, you should not also define the panel's ID in the import file! Should be logical, I guess, because it's in line with everything else during the import, where IDs are always checked first in the file.

PR #115, opened 2016-08-31
Make LOVD more customizable

### LINKED ISSUE #61: Add some features making LOVD more customizable, leading to fewer differences between installations
Within the LUMC, we have a few LOVD3 installations with differences in the code. Let's fix these so we won't need different code bases anymore.
- [x] Add feature disallowing submitter registration
  - [x] Add setting to CONFIG TABLE (inc-sql-tables.php, inc-upgrade.php)
  - [x] Add setting to form, mention the options through messages as well.
  - [x] Made link on template disappear.
  - [x] Make form disable registration (creating a user should still be possible!)
  - [x] Write tests.
- [x] Add feature allowing, between certain dates, a message on top of every page.
  - [x] Messages should have:
    - [x] Start date, End date
    - [x] Message type (info, question, warning, stop)
    - [x] Message body
    - [x] Make LOVD read-only
  - [x] Create table in inc-sql-tables.php and inc-upgrade.php.
  - [x] Add table in inc-init.php.
  - [x] Create object that handles this data type.
  - [x] getForm(), checkFields().
  - [x] VL, VE.
  - [x] Create link in setup.php and for Setup tab.
  - [x] Make template display the messages.
  - [x] Enforce read-only setting.
    - [x] Block login for lower than manager and end current sessions.
    - [x] Block submitter register.
    - [x] Decided for now to not force disallowing edits. Also no warnings will be shown on forms; Managers will now best what (not) to do, considering the announcement that they made themselves.
  - [x] Write tests.
- [x] Describe these features in the manual.
  - [x] New setting to disable submitter registration.
  - [x] Subsection or chapter on messages.

ifokkema:
Implemented the setting to block submitter registration.
- Setting is configured at installation time, default is 1 (on).
- Setting can be turned on and off at will.
- User can not register when setting is turned off. User is suggested to go elsewhere, and a link to search for alternative databases is provided.
- Link from header is removed if setting is off.

Implemented announcements.
- Announcements can be created, edited, and deleted.
- When announcements set LOVD to be read-only, all users lower than Managers are logged out.
- The login form shows a warning that only Managers and higher can log in.
- When a lower level user tries to log in when LOVD is in read-only state, an appropriate error message is given.
- The "register as submitter" link is removed when LOVD is set to be read-only.
- Clarified on the announcement data entry form that true read-only state is not enforced.
- Managers and higher can click on an announcement to go straight to the VE of the announcement in question.

Created tests for submitter registration setting and announcements.
- Also removed email address of previous employee from tests and replaced it with clear, non-personal LOVD email addresses.

Described the new options in the manual, and added chapter "Updating LOVD".
Also updated the changelog, and the documentation page.

HTML version of manual still needs to be created.

Closes #61.


src/inc-init.php
```diff
@@ -476,6 +476,10 @@
 if ($_CONF = $_DB->query('SELECT * FROM ' . TABLE_CONFIG, false, false)) {
     // Must be two-step, since $_CONF can be false and therefore does not have ->fetchAssoc().
     $_CONF = $_CONF->fetchAssoc();
+    if ($_CONF) {
+        // See if the database is read-only at the moment, due to an announcement with the configuration.
+        $_CONF['lovd_read_only'] = (bool) $_DB->query('SELECT COUNT(*) FROM ' . TABLE_ANNOUNCEMENTS . ' WHERE start_date <= NOW() AND end_date >= NOW() AND lovd_read_only = 1')->fetchColumn();
```
mkroon1: It seems a bit strange to check for the read-only status this way, compared to having the field `lovd_read_only` part of the config table.

On the other hand, this way the read-only status is terminated automatically when the announcement ends. So I guess we could leave it like this.

mkroon1:
I generated the HTML manual. Ready to merge, I say.

src/inc-init.php
ifokkema: It's either this, or me updating the `$_CONF` setting every time to reflect the announcement's settings :wink: 
This is, I guess, the lesser of two evils.

PR #120, opened 2016-09-05
Fix proxy server problems

### LINKED ISSUE #112: SOAP using HTTPS over proxy using PHP 5.6.0 doesn't work
Edits have been made already in the SOAP library and LOVD's connectivity function.
- [x] Perform final tests, have user test last two scenarios.
- [ ] Fix SOAP client after results come in.
- [ ] Commit to branch, let others perform code review.

ifokkema:
Fixed bug in the proxy server usage.
- The proxy server settings were often ignored by LOVD, and proxy servers could not be used when using HTTPS, such as with Mutalyzer resources.
- SNI_server_name is deprecated in PHP >= 5.6.0, and is replaced by peer_name. This setting however, can not successfully be used before that version, it seems. Changed this setting for 5.6.0 and up only, in inc-lib-init.php. The SOAP client library in class/soap_client.php drops SNI and peer verification completely.
- lovd_getUDForGene() was ignoring the proxy settings.
- SOAP client now turns WSDL cache off, which slows down the requests a bit, but make sure the call will actually work even in difficult situations.

Added script that tests connectivity to external resources.
- This is not for general release, but is used to send to users having issues with network connectivity.

Removed temp_tests, they are all already performed in the admin testing.
- Also unvoluntairy whitespace changes applied by the IDE.

Closes #112.


mkroon1:
:+1: @ifokkema: will you merge?

PR #130, opened 2016-09-16
Fixing security issues with F&R

ifokkema:
- Fixed SQL injection security issue with F&R feature.
- 	Fixed possible XSS security problem in the F&R form.


ifokkema:
@mkroon1: Unfortunately, I had to delay the release once more. I had noticed a query error in F&R, that apparently was an SQL injection security issue, and I also noticed the possibility of JS injection/XSS. I took care of those, but I need your knowledge of the code to make sure I haven't missed anything. So let's look at it on Monday, and decide on further steps (release or fix more).

PR #133, opened 2016-09-22
Implemented tests for checking the custom links.

ifokkema:
- Added test for checking the custom links.
- Also implemented login() and logout() methods in the tests, and applied them everywhere.
  - Because we're no longer waiting 10 seconds each log out, this should also speed up the tests.
- Also updated broken custom links in all data import files and test files.


mkroon1:
We should revert 9752063 before merging to master, it turns off most of the tests and takes screenshots which are unnecessary (I think).

ifokkema:
>We should revert 9752063 before merging to master, it turns off most of the tests and takes screenshots which are unnecessary (I think).

No worries, that one has already been reverted by 27898b67 and 627990e7 :)

PR #138, opened 2016-09-29
Fix/test fnr

mkroon1:
Not all peculiarities with the find&replace test are fixed, but with the fallback for the StaleElementReferenceException and some extra sleeps, it seems to be more stable now.

@ifokkema: please notice the introduced RefreshingWebElement which can act as a normal [WebDriverElement](http://facebook.github.io/php-webdriver/classes/WebDriverElement.html) but will catch StaleElementReferenceException and retry to find the element and execute the action.


mkroon1:
@ifokkema: please look at changes.

PR #145, opened 2016-11-11
Merge LOVD+ updates into LOVD3

ifokkema:
Included many, many optimizations and minor changes from LOVD+. Most changes actually don't affect LOVD3, but are committed here to make the differences in the code base smaller.
- The code bases of LOVD3 and LOVD+ are brought closer together.
- Introduced the LOVD_plus constant, which tells the code base whether this is an LOVD3 or LOVD+ installation.
- Some changes specifically for LOVD+ are introduced in the LOVD3 code base, only active when LOVD_plus is set to true.
- These changes are mostly cosmetic, but decrease the differences between LOVD3 and LOVD+, making it easier to share improvements.
- Introduced changes:
  - Minimum user level needed for submitting data or seeing non-public data is now configurable, and is different for LOVD+.
  - LOVD+ has a different logo, and has submitter registration disabled by default.
  - Introduced new PDO method fetchAllGroupColumn(), that returns an associative array, grouping the results on the first column.
  - LOVD+ disables prettifying Age fields.
  - Added a hidden column to the Logs VL that can search the log entries.
  - Add additional event names (LOVD+ only) that introduce links in the log entries.
  - Adapted user form to explain additional LOVD+ user level, if LOVD+ is active.
  - Fixed typos in Columns data entry form.
  - LOVD+ does not check VOT entries for determining the DBID, and does not verify the DBID after its generation, to speed up imports.
  - LOVD+ does not update gene's timestamps after adding, editing, or importing entries.
  - LOVD+ has different default pages to redirect to.
  - LOVD+ does not send out emails when data is added or edited.
  - LOVD+ does not show screenings below Individuals VEs.
  - Added rewrite conditions for LOVD+ files.
  - LOVD+ doesn't allow individuals with an identical Lab-ID.
  - LOVD+ doesn't allow assigning individuals to panels.
  - LOVD+ doesn't show VOT specific variant effect fields.
  - Merged LOVD+ menu into LOVD3 menu.
  - LOVD+ allows downloading of gene panels, but doesn't allow any other downloads.
  - LOVD+ prefers gene panels over the genes VL.
  - LOVD+ doesn't assign curators to genes.
  - LOVD+ does not show information on regular LOVD3 updates.
  - LOVD+ reduces the number of columns in the Genes VL to speed up the VL.
  - LOVD+ hides the "Updated" fields in the Genes VE since it's not updating that field.
  - LOVD+ adds a column with transcript info to the Genes VL for the Gene Panel edit form.
  - LOVD+ adds columns on the curation status and confirmation status on the Variants VE.
  - Close the session when importing, to not lock it.
  - LOVD+ uses only one Gene object for the checkFields() and for the columns list, to not require so much memory.
  - LOVD+ imports variants as non-public by default.
  - LOVD+ mentions the file name that was imported, in the log entry related to the import.
  - LOVD+ uses requireAuth() also for authorization on analysis runs and analyses (based on the screening).
  - LOVD+ has also data paths as configuration values; add the code that checks the directories.
  - Missing INI settings now no longer cause notices; they will be set to false.
  - Added lovd_convertSecondsToTime(), a function that formats a time string based on the number of seconds.
  - Added menu icons from LOVD+, perhaps useful for LOVD3 as well.
  - The progress bar now respects the format setting, allowing it to output plain text as well.
  - The import page will not output JS if the output format is not set to text/html.
  - lovd_errorPrint() now respects the format setting, allowing it to output in plain text as well.
  - lovd_showInfoTable() now respects the format setting, allowing it to output in plain text as well.
  - The jeegoo menu now has a shorter delay time.
  - The jeegoo menu now allows for submenus.
  - Locus group "other" is no longer fully ignored.
  - LOVD+ during installation also registers analysis statuses, curation statuses and confirmation statuses.
  - LOVD+ during installation also creates an index on the VariantOnGenome/DBID column.
  - Sync the styles of selection lists and textareas in a data table with the input fields.
  - Added some styles for the selection lists.
  - Added PubMed article source and Alamut custom link.
  - Make the allow_submitter_registration depend on the LOVD_plus constant.
  - Dropped old table constants that were no longer in use.			
  - Added LOVD+ code for showing analyzer information.
  - Let LOVD+'s analysis_by and analysis_approved_by columns be handled similar to other user ID columns.
  - Let the query parser handle table aliases longer than 3 characters.
  - Remove tags from field names when we format the legend and the search field's title attribute.
  - Remove download options for VLs for LOVD+.
  - Adding and removing diseases from individuals is now logged.
  - LOVD_Individuals::checkFields() required $zData to be passed.
  - Added menu options and additional authorization checks for LOVD+ only.			
  - Added images for jeegoocontext submenus.			
  - Fixed jeegoocontext styles for longer menu items and fixed image URLs.			
- 	Fixed notices during tests; made it prefill the LOVD_plus constant.
- Larger changes:
  - Rewrote data access methods in LOVD_Object to deal with revision and linking tables.
  - deleteEntry(), getCount(), insertEntry(), loadEntry(), updateEntry() and viewEntry() were rewritten to deal with revision and linking tables.
  - updateEntry() will now calculate a diff between the data in the database, and not update if nothing changed.
  - Check "raw" WHERE and HAVING statements for the filter ID, as they can contain more than just what's given through the VL.


ifokkema:
@mkroon1: I know this thing is huge, but if you could scroll through it, then you at least have an idea of the changes done. Some changes (objects.php) are major, most are minor.

PR #147, opened 2016-11-11
Fixed problem with sticky header when announcements are present.

ifokkema:
- When announcements were present, the sticky header would "float" away from the top of the page when scrolling.
- Now, the announcements are grouped in a DIV together with the header, to have them float together on the top of the page.
- The new offset for the table with the menu is now calculated based on assumptions of the announcement height in the HTML, but then calculated by JS directly after rendering the header tables, after which the margin is corrected if needed by the actual measurements.

### DIFF
--- a/src/class/template.php
+++ b/src/class/template.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2012-03-27
- * Modified    : 2016-11-11
+ * Modified    : 2016-11-16
  * For LOVD    : 3.0-18
  *
  * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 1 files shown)


src/class/template.php
```diff
@@ -782,6 +787,12 @@ function lovd_changeURL ()
               '</TABLE>' . "\n\n" .
               '<IMG src="gfx/trans.png" alt="" width="' . $nTotalTabWidth . '" height="0">' . "\n\n");
 
+        // Measure the height of the sticky header (can depend on announcements or
+        // font settings and such), and adapt the menu table to have a margin of this height.
+        print('<SCRIPT type="text/javascript">' . "\n" .
+              '  $("table.logo :eq(1)").css("margin-top", $("#stickyheader").height() + "px");' . "\n" .
```
mkroon1: Maybe it's best to use outerHeight() instead of height(), that includes padding and border (if that's ever added to the header div).

ifokkema: Good idea! That would be `outerHeight(true)` then, to include the margin as well. Would you like to make this change, or shall I do it?

mkroon1:
Changed it to `outerHeight(true)`. It's now also only printed when there are actually announcements.

PR #154, opened 2016-11-18
Let users configure an API token

ifokkema:
Users can now create their own API tokens to use for the future LOVD data submission API.
- Adapted data model to store authorization token and its expiration date.			
- Added fields to users VE; the token (hidden, only link shown for more info) and the token expiration.
- Needed `lovd_convertSecondsToTime()` from LOVD+ for the latter.
- Clicking the link for the API token information opens a dialog that's fed through an AJAX request.
- Dialog shows some information on the token, also displays buttons to create and revoke the token.
- Dialog displays when token has expired, and asks for confirmation when creating a new one while one is still valid.
- Dialog can create a new token, setting the expiration.
- Dialog can revoke a token.
- Both features are CSRF protected.
- Modernized `ajax/viewentry.php` and enabled it to load Users VEs. Also had to change `lovd_isAuthorized()` for that.
- Put a DIV around the Users VE so we can replace it. Users VE is reloaded when tokens are created or revoked.
- New tokens and revokes are logged.
- Added changelog entry.

### DIFF
--- a/src/class/object_users.php
+++ b/src/class/object_users.php
@@ -487,7 +487,7 @@ class LOVD_User extends LOVD_Object {
                 // Either we're viewing ourselves, or we're manager or up.
 
                 // Auth token links. We don't show the token by default.
-                $zData['auth_token_'] = '(<A href="#" onclick="lovd_showAuthTokenDialog(); return false;">Show / More information</A>)';
+                $zData['auth_token_'] = '(<A href="#" onclick="$.get(\'ajax/auth_token.php/' . $zData['id'] . '?view\'); return false;">Show / More information</A>)';
                 if ($zData['auth_token_expires']) {
                     $tDiff = strtotime($zData['auth_token_expires']) - time();
                     $sDiff = lovd_convertSecondsToTime(abs($tDiff));
--- a/src/class/object_users.php
+++ b/src/class/object_users.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-21
- * Modified    : 2016-11-18
+ * Modified    : 2016-12-09
  * For LOVD    : 3.0-18
  *
  * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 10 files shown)


src/class/object_users.php
```diff
@@ -480,9 +482,21 @@ function prepareData ($zData = '', $sView = 'list')
...
+                // Either we're viewing ourselves, or we're manager or up.
+
+                // Auth token links. We don't show the token by default.
+                $zData['auth_token_'] = '(<A href="#" onclick="$.get(\'ajax/auth_token.php/' . $zData['id'] . '?view\'); return false;">Show / More information</A>)';
+                if ($zData['auth_token_expires']) {
+                    $tDiff = strtotime($zData['auth_token_expires']) - time();
+                    $sDiff = lovd_convertSecondsToTime(abs($tDiff));
+                    $zData['auth_token_expires_'] = '<SPAN title="' . $zData['auth_token_expires'] . '">' . ($tDiff > 0? 'In ' . $sDiff : 'Expired ' . $sDiff . ' ago') . '</SPAN>';
+                } else {
+                    $zData['auth_token_expires_'] = (!$zData['auth_token']? '' : '- (Never)');
```
mkroon1: Perhaps it's clearer if the expiration is displayed as "Never" instead of "- (Never)".

Another idea is to show everything related to the token in one VE row, like:
*401b30e3b8b5d629635a5c613cdb7919 (expires Never) (more information / revoke)*
Where *more information / revoke* links to the dialog to manage the token.

src/class/object_users.php
```diff
@@ -480,9 +482,21 @@ function prepareData ($zData = '', $sView = 'list')
...
             if (lovd_isAuthorized('user', $zData['id']) === false) {
                 // Not authorized to view hidden data for this user; so we're not manager and we're not viewing ourselves. Nevermind then.
-                unset($this->aColumnsViewEntry['ownes_']);
+                unset($this->aColumnsViewEntry['ownes_'], $this->aColumnsViewEntry['auth_token_'], $this->aColumnsViewEntry['auth_token_expires_']);
             } else {
-                // Either we're viewing ourselves, or we're manager or up. Like this is easy, because now we don't need to check for the data status of the data.
+                // Either we're viewing ourselves, or we're manager or up.
+
+                // Auth token links. We don't show the token by default.
+                $zData['auth_token_'] = '(<A href="#" onclick="$.get(\'ajax/auth_token.php/' . $zData['id'] . '?view\'); return false;">Show / More information</A>)';
```
mkroon1: When the AJAX request fails, nothing is shown to the user, this may be confusing.

src/class/object_users.php
ifokkema: > Perhaps it's clearer if the expiration is displayed as "Never" instead of "- (Never)".

I was doubting about that, but chose the latter. I don't think I have much of a preference, so if you prefer "Never" we can change it :wink: 


> Another idea is to show everything related to the token in one VE row, like:

We shouldn't show the token in the VE directly. Even though it's an MD5, it's still sensitive information and should, in my opinion, only be shown when really requested.

But yes, we could merge the two, if you prefer that.

src/class/object_users.php
ifokkema: > 404, when the client has no network connection, or when there's a timeout.

I guess failing network connection is the only real thing we would deal with, and I guess a proper message is OK even though the entire LOVD would stop to work at that time :wink: 


> We could have a simple funciton (...) a loading icon (...) and a warning icon when it fails.

That's very much overkill. This is a local (also LOVD) script that returns instantly, and it doesn't use external tools. I once happened to implement a waiting icon for such a local script, and it didn't even show because the connection was so fast (remember that showing the loading icon is yet another request, so it will always be late). Also since the chance of failing is very much zero, and if it would fail, so would the rest of LOVD, so the failure icon wouldn't load either. A simple alert that errors is more than enough, I'd say...

PR #158, opened 2016-12-01
Feat/convert lovd2

### LINKED ISSUE #96: Include script to convert the LOVD2 download format to the LOVD3 format
LOVD3 still lacks an importer for LOVD2 data. While this is a pretty complicated task, we can relatively easily start with a conversion tool that takes an LOVD2 file and convert it to LOVD3 format. It will not take care of transferring user accounts, it will not take care of mapping custom columns, but at least it will make moving data a lot easier. It may be (loosely) based on the `get_variant_info.php` script developed by @ifokkema that is fetching information needed for LOVD3 using limited (sort of LOVD2 style) data and fetches transcript IDs etc.
- [x] This script will handle one gene per conversion only.
- [x] The gene must exist in the LOVD3 installation already.
- [x] Before conversion is done, the transcript should be chosen. This transcript should already exist in LOVD3.
- [x] The script should take the LOVD2 format, and create the full LOVD3 format.
- [x] Information missing in the LOVD2 file that LOVD3 needs should be provided by this script (`get_variant_info.php` should take care of all of that, I think).

### LINKED ISSUE #155: Undefined index notice in LOVD_Custom::checkFields()
>  **Notice: Undefined index: Phenotype/Age/Onset in src/class/object_custom.php on line 375**

Thrown when importing phenotype records not containing the 'Phenotype/Age/Onset' field.

mkroon1:
Added script for conversion of LOVD2 export format to LOVD3 import format.

Fixes #96 
Fixes #155


src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1312 @@
...
+
+
+        // Create new disease if necessary.
+        $aDisease = lovd_getRecordForHeaders($aOutputHeaders['disease'], $aRecord,
+            $aSections['disease']);
+        if (!lovd_empty($aDisease['name'])) {
+            list($aDisease['id'], $bCreateNewDisease) = lovd_getDiseaseID($aDisease['name']);
+            if ($aDisease['id'] === false) {
+                // Something went wrong determining the disease, stop parsing.
+                break;
```
ifokkema: Why stop parsing here? It would be more useful to continue, so the user will see all the errors in the entire file, instead of having to solve the problems one by one and retrying the conversion each time.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1312 @@
...
+        }
+
+        // Create VOG/VOT records
+        $aVOGRecord = lovd_getRecordForHeaders($aOutputHeaders['vog'], $aRecord,
+            $aSections['vog']);
+        $aVOGRecord['chromosome'] = $zTranscript['chromosome'];
+        if ($aVOGRecord['edited_by'] === false) {
+            // No curator ID was available, set submtiter ID.
+            $aVOGRecord['edited_by'] = lovd_convertUserID($sSubmitterID, 'submitter');
+        }
```
ifokkema: What if the data simply had never been edited? The `edited_by` would be null, and should remain `null`...

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1312 @@
...
+                $aIndividual = lovd_getRecordForHeaders($aOutputHeaders['individual'], $aRecord,
+                    $aSections['individual']);
+                if ($aIndividual['created_by'] === false) {
+                    // No curator ID was available, set submtiter ID.
+                    $aIndividual['created_by'] = lovd_convertUserID($sSubmitterID, 'submitter');
+                }
+                if ($aIndividual['edited_by'] === false) {
+                    // No curator ID was available, set submtiter ID.
+                    $aIndividual['edited_by'] = lovd_convertUserID($sSubmitterID, 'submitter');
+                }
```
ifokkema: What if the data simply had never been edited? Then `edited_by` would be null, and should remain `null`...

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1312 @@
...
+
+    $aFixedUserInfos = array(
+        array('submitter', 'submitterid_fixed', 'sFixedSubmitterID', $sFixedSubmitterIDInput),
+        array('curator', 'curatorid_fixed', 'sFixedCuratorID', $sFixedCuratorIDInput),
+    );
+
+    foreach ($aFixedUserInfos as $aFixedUserInfo) {
+        list($sIDtype, $sFormField, $sGlobalVar, $sInput) = $aFixedUserInfo;
+        if (ctype_digit($sInput)) {
+            $GLOBALS[$sGlobalVar] = $sInput;
```
ifokkema: Why this way of referring to the variable? Why not directly?

src/scripts/convert_lovd2.php
mkroon1: I agree, that was the intention when I started on this, will fix

src/scripts/convert_lovd2.php
mkroon1: I didn't want to duplicate the code for setting the variables for submitter and curator (fixed IDs and translation tables)

src/scripts/convert_lovd2.php
mkroon1: On second thought, `$aIndividual['edited_by'] === false` holds only when the input field was "0", set by `lovd_convertUserID()`. Empty edited_by fields will remain empty in the output.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+                list(, $sFieldname) = explode('/', $sHeader, 2);
+                foreach ($aSectionIDs as $sSection) {
+                    $aSection = $aSections[$sSection];
+                    if (isset($aSection['customcol_prefix'])) {
+                        $sNewFieldName = $aSection['customcol_prefix'] . '/' . $sFieldname;
+                        if (in_array($sNewFieldName, $aSection['db_fields'])) {
+                            // Set output header with new LOVD3 prefix (e.g. Individual).
+                            $aOutputHeaders[$sSection][$i] = $sNewFieldName;
+                            continue 2;
+                        }
```
ifokkema: In this block, you make a new field name for each column with *every* section. Shouldn't you be able to see which section this column belongs to already, by looking at the original prefix?

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+                // Try to find default custom column translation.
+                foreach ($aCustomColLinks as $sPrefix => $aCustomColDefault) {
+                    if (strpos($sHeader, $sPrefix) === 0) {
+                        list($sSection, $sPrefixOut) = $aCustomColDefault;
+                        $sHeaderOut = str_replace($sPrefix, $sPrefixOut, $sHeader);
+                        $aOutputHeaders[$sSection][$i] = $sHeaderOut;
+                        $_WARNINGS[] = 'Warning: linked "' . $sHeader . '" to non-existing ' .
+                            'column "' . $sHeaderOut . '" in output section "' .
+                            $aSections[$sSection]['output_header'] . '"';
+                        continue 2;
```
ifokkema: This warning is triggered all the time for columns that do exist in the database. I can't debug it myself, there's very little comments and I'm not quite following it.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+    // are defined in the upload form.
+    global $sFixedCuratorID, $aCuratorTranslationTable, $_WARNINGS;
+
+    $LOVD2UserIDClean = intval($nLOVD2UserID);
+    // Convert curator ID.
+    if ($LOVD2UserIDClean === 0) {
+        // '0' in LOVD2 export means the submitter ID should be used.
+        // Return false.
+        return false;
+    }
```
mkroon1: In 86895d8, I've avoided the use of `intval()` so that there are no `0`s introduced there. If the input for a user ID field is `0`, this function returns `false`, downstream code should convert that based on the submitter ID.

Note that in the 86895d8 version, when e.g. input field `ID_variant_created_by_` is empty, it will be untouched by this function. However, since it is linked to the vog `created_by` field, `lovd_getRecordForHeaders()` will set that field to `0` (the default value defined in `$aFieldLinks`). It will then **not** be converted again based on the submitter ID. I think this is correct behavior.

src/scripts/convert_lovd2.php
mkroon1: It *tries* to make a link for the field with every section, but stops when it finds a field name with the same name (excluding prefix). The original prefix and the section prefix may be different, so I don't see how to simplify this.

src/scripts/convert_lovd2.php
mkroon1: Fixed this by wrapping it with a check if the DB field exists.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+    foreach ($aOutputHeaders as $nInputIdx => $sHeader) {
+        if (is_int($nInputIdx) || ctype_digit($nInputIdx)) {
+            // Numeric key $nInputIdx defines link to field in input record $aRecord.
+            if (!empty($aNewRecord[$sHeader]) && !empty($aRecord[$nInputIdx])) {
+                $_WARNINGS[] = 'Warning: doubly-linked field already has a value "' .
+                               $aNewRecord[$sHeader] . '", alternate value will get lost: "' .
+                               $aRecord[$nInputIdx] . '"';
+                continue;
+            }
+            $aNewRecord[$sHeader] = $aRecord[$nInputIdx];
```
mkroon1: Agreed. Fixed this in d6c2d6f where the target field is made sure to be empty.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+
+                // Create screening2gene record.
+                $aScreening2Genes[] = array('screeningid' => $nScreeningID,
+                                            'geneid' => $zTranscript['geneid']);
+
+                // Create phenotype record.
+                $aPhenotype = lovd_getRecordForHeaders($aOutputHeaders['phenotype'], $aRecord,
+                    $aSections['phenotype']);
+                // Skip phenotype because there is no data in phenotype
+                // record except for ID fields.
```
mkroon1: > What is this comment referring to?

Something either moved or not implemented yet, I suspect. I've implemented this in 1b458f5

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+        $oResponse = lovd_callJSONService($sMappingURL);
+        if ($oResponse && !isset($oResponse->errorcode) && !isset($oResponse->faultcode)) {
+            $aVOGRecord['position_g_start'] =   min($oResponse->start_g, $oResponse->end_g);
+            $aVOGRecord['position_g_end'] =     max($oResponse->start_g, $oResponse->end_g);
+            $aVOGRecord['type'] =               $oResponse->mutationType;
+            $aVOTRecord['position_c_start'] =   $oResponse->startmain;
+            $aVOTRecord['position_c_start_intron'] =    $oResponse->startoffset;
+            $aVOTRecord['position_c_end'] =             $oResponse->endmain;
+            $aVOTRecord['position_c_end_intron'] =      $oResponse->endoffset;
+        }
```
mkroon1: :+1:

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1318 @@
...
+    print('<H3>Conversion log:</H3>
+    <TEXTAREA id="header_log" cols="100" rows="10" style="font-family: monospace; 
+        white-space: nowrap; overflow: scroll;">' .
+        implode("\n", $_WARNINGS) .
+    '</TEXTAREA><BR><BR>');
+
+    if (lovd_error()) {
+        print('<B>There were fatal errors during conversion:</B>');
+        lovd_errorPrint();
+    }
```
mkroon1: > I think you can't get here if you have lovd_error() returning true. I can't be sure, as lovd_errorAdd() is called in 5 different functions, and there's no way for me to see in a structured way how they are calling each other.

Do you mean you can't see a call to `lovd_errorAdd()` that would trigger this print statement? `lovd_getDiseaseID()` can add errors during conversion if the search for a disease name is ambiguous.

> If you can't get here, this part can go. If you can get here, the bottom part shouldn't be displayed if lovd_errorAdd() is true.

I thought the conversion output might be useful somehow. What else would you want displayed, the submission form?

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1346 @@
...
+    'Patient/Times_Reported' =>         array('individual', 'panel_size'),
+    'Patient/Phenotype_2' =>            array('phenotype',  'Phenotype/Additional'),
+    // Note that 'Patient/Phenotype/Inheritance' automatically maps to Phenotype/Inheritance as well.
+    'Patient/Occurrence' =>             array('phenotype',  'Phenotype/Inheritance',        'lovd_convertInheritance'),
+    'Patient/Mutation/Origin' =>        array('vog',        'VariantOnGenome/Genetic_origin',   'lovd_convertOrigin'),
+    'ID_pathogenic_' =>                 array('vog',        'effectid'),
+    'ID_status_' =>                     array('vog',        'statusid',                     'lovd_convertStatus'),
+    'ID_variant_created_by_' =>         array('vog',        'created_by',                   'lovd_convertCuratorID'),
+    'variant_created_date_' =>          array('vog',        'created_date'),
+    'ID_variant_edited_by_' =>          array('vog',        'edited_by',                    'lovd_convertCuratorID'),
```
ifokkema: Just agreed to this change:
- The `edited_by` should not be overwritten with the fixed curator ID, unless the `edited_by` is actually filled in with something in the LOVD2 file.

An non-empty `edited_by` field results in the entry being imported in LOVD3 as "edited" whereas in LOVD2 it has never been edited.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1346 @@
...
+    'variant_edited_date_' =>           array('vog',        'edited_date'),
+    'ID_patient_created_by_' =>         array('individual', 'created_by',                   'lovd_convertCuratorID'),
+    'patient_created_date_' =>          array('individual', 'created_date'),
+    'ID_patient_edited_by_' =>          array('individual', 'edited_by',                    'lovd_convertCuratorID'),
+    'patient_edited_date_' =>           array('individual', 'edited_date'),
+    'ID_patientid_' =>                  array('individual', 'id',                           'lovd_autoIncIndividualID'),
+    'ID_variantid_' =>                  array('vog',        'id',                           'lovd_autoIncVariantID'),
+    'ID_allele_' =>                     array('vog',        'allele'),
+    'ID_submitterid_' =>                array('vog',        'owned_by',                     'lovd_convertSubmitterID'),
```
ifokkema: Just agreed to these changes:
- The `owned_by` field should not be filled in from the static curator ID directly, only perhaps indirectly (see below).
- The `owned_by` field should contain the translated `ID_submitterid_` field, the static submitter ID, or otherwise the *translated* value that is in the `created_by` field. So it's best to apply the static curator ID first to the `created_by`, before you fill the `owned_by` field.

src/scripts/convert_lovd2.php
```diff
@@ -0,0 +1,1346 @@
...
+        array('Fixed submitter ID', 'All user IDs in the imported file will be replaced with ' .
+            'this value. (E.g. all IDs in the edited_by, created_by fields)', 'text',
+            'submitterid_fixed', 10),
+        array('', '', 'note', 'All user IDs in the imported file will be replaced with this ' .
+            'value. (E.g. all IDs in the edited_by, created_by fields)'),
+        array('Submitter ID translation table', '', 'textarea', 'submitterid_translation', 20, 6),
+        array('', '', 'note', 'Translation table for user IDs. On every line an LOVD2 user ID ' .
+            'is expected, followed by an LOVD3 user ID, separated by whitespace. Any submitters ' .
+            'present in the selected file will translated according to this table.'),
+        array('Fixed curator ID', '', 'text', 'curatorid_fixed', 10),
```
ifokkema: It would be good to let this field have some explanation too, like the fixed submitter ID.

src/scripts/convert_lovd2.php
mkroon1: Fixed in 1949a03 (note that the commit message is off -- the fixed user ID is only ignored for `edited_by` fields)

src/scripts/convert_lovd2.php
mkroon1: Fixed in 5d83fd7

src/scripts/convert_lovd2.php
mkroon1: Fixed in b3adf65

PR #160, opened 2016-12-08
VarioML JSON API for direct submissions

ifokkema:
- Also contains all code from the feat/add_API_token and the feat/getVariantInfo branches.
- Added LOVD_API class to take over basic API functionality.
  - The LOVD_API class verifies the Accept header and picks the proper format, and takes over the URL verification from the api.php file for the LOVD2-style API.
  - The LOVD_API class is capable of outputting JSON or plain text, and makes sending out HTTP headers and output easier.
- Implemented the /api/v#/submissions URL.
  - LOVD_API class checks for a version in the URL, and checks it.
  - LOVD_API class checks the given resource in the URL, and verifies if we can handle it.
  - LOVD_API class checks the currently used method for the given resource.
  - LOVD_API class parses out the ID (currently unused) and checks if the rest of the URL is empty as it should be.
  - LOVD_API class makes sure we don't POST to a resource ID.
  - sendHeader() also sends the Allow header, if needed.
- Create new class to submissions.
  - LOVD_API calls its own method for handling POST requests.
  - LOVD_API::processPOST() then checks the resource type to figure out what to do.
  - Created LOVD_API_Submissions class to handle the submissions resource.
  - LOVD_API throws an HTTP 500 if no handler picks up the request.
  - LOVD_API_Submissions checks for received data, and checks for a maximim size.
  - LOVD_API_Submissions checks for the Content-type being sent by the client, and for the '{' in the JSON.
  - LOVD_API_Submissions tries to decode the JSON, and gracefully handles the error otherwise.
  - All methods in the resource handling class return false on failure, which is passed down to the API class which then sends the prepared output.
- JSON is cleaned now; comments removed and layers added where needed.			
- VarioML file is getting parsed and checked (missing data, values should be valid, auth tokens, etc).
  - HTTP 401 is sent when token is incorrect. WWW-Authenticate header is sent, too, to comply to the HTTP specs.
  - RefSeqs are looked up in the database, the closest matching is picked.
  - If there are none available for a variant but requested, then we give a hard error. If genes were given, and available in the database, we list alternatives.
  - Otherwise, we remove the ones that don't match, and warn for those, mentioning which transcripts we will be using.
- Allow for data paths to be configured, where the API will dump the import files.
- The LOVD API will throw a HTTP 500 with a clear error message with the admin contact details if there is no data file path configured.
- convertVarioMLToLOVD() creates an LOVD data array based on the VarioML data.
- The variants get their position fields filled in.
- The API now writes data to an import file in the configured data path.
- The VariantOnTranscript/Exon field is now no longer mandatory by default, allowing the submission API to directly submit data, since it doesn't contain the Exon field.
- Added changelog entries.
- lovd_writeLog() is now able to log for other users.


src/ajax/viewentry.php
```diff
@@ -43,19 +48,26 @@
...
-if ($_GET['object'] == 'Transcript_Variant') {
-    list($nVariantID, $nTranscriptID) = explode(',', $_GET['id']);
+if ($sObject == 'Transcript_Variant') {
+    list($nVariantID, $nTranscriptID) = explode(',', $nID);
     lovd_isAuthorized('variant', $nVariantID);
+} elseif ($sObject == 'User') {
+    lovd_isAuthorized(strtolower($sObject), $nID);
+    // Users viewing their own profile should see a lot more...
+    if ($_AUTH['id'] == $nID && $_AUTH['level'] < LEVEL_CURATOR) {
+        $_AUTH['level'] = LEVEL_CURATOR;
```
mkroon1: I don't like this (setting a fake user level to 'trick' the rest of the code). But I don't have an alternative, at least not a simple one, to get the same result.

src/class/api.php
```diff
@@ -0,0 +1,356 @@
...
+                header('HTTP/1.0 501 Not Implemented');
+                exit;
+            }
+
+            // Parse URL to see what we need to do.
+            list(,,
+                $this->sResource, // 2
+                $this->sGene,     // 3
+                $this->nID) = $_PE;
+
```
mkroon1: It might be an idea to move the unpacking of $_PE to line 141, so indexing $_PE is not necessary anymore in the code above.

src/class/api.php
```diff
@@ -0,0 +1,356 @@
...
+                    $this->aResponse['errors'][] = 'Requested version is unavailable.';
+                    $this->sendHeader(400, true); // Send 400 Bad Request, print response, and quit.
+                } else {
+                    $this->nVersion = (int) $aRegs[1];
+                }
+                array_shift($aURLElements); // Take the version off.
+            }
+
+            // Next, should be resource.
+            $this->sResource = array_shift($aURLElements);
```
mkroon1: I think `$this->sResource` has been set at line 166, right?

src/class/api.php
```diff
@@ -0,0 +1,356 @@
...
+                $this->aResponse['errors'][] = 'Method not allowed here. Options: ' . implode(', ', $this->aResourcesSupported[$this->sResource]);
+                $this->sendHeader(405, true); // Send 405 Method Not Allowed, print response, and quit.
+            }
+
+            // If we're here, the API regarded the call acceptable, and assigned
+            //  an API version higher than 0 (the LOVD2-style API).
+            // If we're at version 1 or higher, let this new API handle it.
+            // Since each method requires very specific code, the methods are
+            //  handled separately.
+            // These methods should process the request and stop PHP execution.
```
mkroon1: Is the last line in the comment still up to date? It seems that `processPOST()` now always returns

mkroon1:
I'm done reviewing and did not find any blocking problems. Although don't have enough knowledge on the json format to say many intelligent things about all the varioML-related code in api.submissions.php

src/ajax/viewentry.php
ifokkema: Same here... :roll_eyes:

src/class/api.php
ifokkema: No, we can't. The thing is that the LOVD2 and the LOVD3 API are joint still in line 141, and this code is specifically for the LOVD2 API. The LOVD3 API has a different URL schema, and shouldn't be bothered by the LOVD2 API's URL construction.

src/class/api.php
ifokkema: There's an "else" in between there, so those lines of code are never both run :wink:

src/class/api.php
ifokkema: You're right, it was speaking of the previous structure. I've removed the comment, thanks!

ifokkema:
> I'm done reviewing and did not find any blocking problems. Although don't have enough knowledge on the json format to say many intelligent things about all the varioML-related code in api.submissions.php

(non-intelligent comments are also welcome :wink:)
Cool, I'll merge, then! :+1:

PR #161, opened 2016-12-08
Fix/only full group by

### LINKED ISSUE #144: Make LOVD compatible with the `ONLY_FULL_GROUP_BY` MySQL setting.
LOVD is currently not compatible with the `ONLY_FULL_GROUP_BY` MySQL setting.

From the MySQL manual:
> In MySQL 5.7.5, the ONLY_FULL_GROUP_BY SQL mode is enabled by default because GROUP BY processing has become more sophisticated to include detection of functional dependencies. However, if you find that having ONLY_FULL_GROUP_BY enabled causes queries for existing applications to be rejected, either of these actions should restore operation:
> 
> If it is possible to modify an offending query, do so, either so that nonaggregated columns are functionally dependent on GROUP BY columns, or by referring to nonaggregated columns using ANY_VALUE().

mkroon1:
Checked all queries in the code for issue #144. Those that were modified or complex were tested in the application.

Attached file [group_by_occurrences.txt](https://github.com/LOVDnl/LOVD3/files/639784/group_by_occurrences.txt) contains a listing of queries containing GROUP BY or DISTINCT with annotation of what action was performed (see legend in file).

Fixes #144

### DIFF
--- a/src/class/object_custom_viewlists.php
+++ b/src/class/object_custom_viewlists.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2011-08-15
- * Modified    : 2016-10-14
+ * Modified    : 2016-12-07
  * For LOVD    : 3.0-18
  *
  * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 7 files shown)


ifokkema:
What does `U` mean in the file? It's not mentioned in the legend.

src/class/object_custom_viewlists.php
```diff
@@ -766,6 +775,28 @@ private function getCustomColsForCategory ($sCategory)
...
+            $aNamesAggregated = array_map(function ($sName) use ($sSelectQuery) {
+                if (strpos($sName, $sSelectQuery) === false) {
+                    return 'MIN(`' . $sName . '`) AS `' . $sName . '`';
+                }
+                return '';
+            }, $aCustomColNames);
+            return join(', ', $aNamesAggregated);
+        }
+        return '';
+    }
```
ifokkema: Could you walk me through this one? I don't understand why we need all this extra code, so I must be missing something... :sweat_smile:

src/columns.php
```diff
@@ -1410,7 +1410,7 @@ function lovd_checkSubmittedForm ()
             $nPossibleTargets = count($aPossibleTargets);
         } elseif ($sCategory == 'Phenotype') {
             // Retrieve list of diseases which do NOT have this column yet.
-            $sSQL = 'SELECT DISTINCT d.id, IF(CASE d.symbol WHEN "-" THEN "" ELSE d.symbol END = "", d.name, CONCAT(d.symbol, " (", d.name, ")")) FROM ' . TABLE_DISEASES . ' AS d LEFT JOIN ' . TABLE_GEN2DIS . ' AS g2d ON (d.id = g2d.diseaseid) LEFT JOIN ' . TABLE_SHARED_COLS . ' AS c ON (d.id = c.diseaseid AND c.colid = ?) WHERE c.colid IS NULL';
+            $sSQL = 'SELECT DISTINCT d.id, IF(CASE d.symbol WHEN "-" THEN "" ELSE d.symbol END = "", d.name, CONCAT(d.symbol, " (", d.name, ")")), d.symbol, d.name FROM ' . TABLE_DISEASES . ' AS d LEFT JOIN ' . TABLE_GEN2DIS . ' AS g2d ON (d.id = g2d.diseaseid) LEFT JOIN ' . TABLE_SHARED_COLS . ' AS c ON (d.id = c.diseaseid AND c.colid = ?) WHERE c.colid IS NULL';
```
ifokkema: Could you help me out with this one? I don't understand why MySQL would need these to be selected (I know they are in the `ORDER BY`)...

mkroon1:
In the change listing, `U` means **U**nchanged.

src/class/object_custom_viewlists.php
mkroon1: Sorry, it's more complicated than necessary, and now that I see it again I spotted a bug :sweat_smile: 

I've committed a fix in ee972e4 which is easier to read.

src/columns.php
ifokkema: Seems to have no requirement, removing.

PR #163, opened 2016-12-13
Minor fixes

### LINKED ISSUE #140: Position fields during import sometimes left as NULL
During import, the position_c_*_intron fields are sometimes left at NULL. This disrupts the unique variant view, as grouping with variants that have that field set to 0 does not occur.
- [x] Make the import set that field to 0 when left empty.
- [x] Incorporate the function that extracts positions from the DNA field into the import (see issue #58).
- [x] In the upgrade code, write SQL to fix the already present entries. Similar to:

``` sql
update lovd_v3_variants_on_transcripts set position_c_start_intron = 0 where position_c_start is not null and position_c_start_intron is null;
update lovd_v3_variants_on_transcripts set position_c_end_intron = 0 where position_c_end is not null and position_c_end_intron is null;
```

ifokkema:
- When an Administrator creates a new user, the new default level is Submitter.
- Fixed issues with the variant effect fields:
  - The variant effect columns were checked twice while submitting and importing, possibly resulting in repeated error messages.
  - Fixed bug; The variant effect fields didn't have a default value when importing.
- Fixed issues with the position fields:
  - Fixed bug; Imports could set variant start positions that were higher than variant end positions.
  - When variants are imported without position fields or variant type filled in, LOVD will now predict them.
  - Fixed bug; Variants that were imported with empty intron positions, didn't group together with variants that had a 0 filled in.
    - All variants already imported that way, are fixed.
  - Closes #140.

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -35,6 +35,10 @@
    the Exon field.
  * When a Database administrator creates a new user, the level now selected by
    default will be Submitter instead of Manager.
+ * The variant effect columns were checked twice while submitting and importing,
+   possibly resulting in repeated error messages.
+ * Fixed bug; The variant effect fields didn't have a default value when
+   importing.
 
 
 /************************************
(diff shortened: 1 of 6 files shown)


ifokkema:
@mkroon1: I'm going to merge anyway, if you have time when you come back you can still look at the code if you wish :wink:

Issue #167, opened 2017-01-09
Manage transcripts feature doesn't allow addition of transcript anymore


ifokkema:
Since 3.0-18, the "manage transcripts" feature doesn't allow the addition of a transcript anymore. Repeatedly trying may have a positive result, but the problem should be fixed.

PR #179, opened 2017-01-25
Fix "all variants affecting transcripts" VL

### LINKED ISSUE #176: Viewlist entry count differs from what is displayed
In the variants-affecting-transcripts viewlist, LOVD (sometimes?) shows a count for the total number of entries that is different from the entries that is actually shown. See image below for example:

lovd_bug_20170124

(Reported by Julia)

ifokkema:
Fixed bug; The "all variants affecting transcripts" view showed only one entry.
- This error is related to #167.
- This was caused by the recent changes making LOVD compatible with the new default `ONLY_FULL_GROUP_BY` MySQL setting.
- There was no more GROUP BY defined, after which MySQL grouped all results into one.
- Closes #176: "Viewlist entry count differs from what is displayed".
- Checked all other Custom VLs for this error as well, now they are all fine.

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -13,6 +13,10 @@
    "ONLY_FULL_GROUP_BY" MySQL setting.
    Closes #167: "Manage transcripts feature doesn't allow addition of transcript
    anymore".
+ * Fixed bug; The "all variants affecting transcripts" view showed only one
+   entry. This was caused by the recent changes making LOVD compatible with the
+   new default "ONLY_FULL_GROUP_BY" MySQL setting.
+   Closes #176: "Viewlist entry count differs from what is displayed".
 
 
 /**************************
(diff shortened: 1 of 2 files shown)


mkroon1:
> There was no more GROUP BY defined, after which MySQL grouped all results into one.

I had to think a bit about what was going on here, since I didn't remember removing a GROUP BY statement there. But of course GROUP BY was never there, adding aggregating functions by default for all custom columns caused this effect.

In the future we may want to change the approach by only adding aggregating functions if there's actually a GROUP BY statement defined. For now, this fix seems to do the trick.

ifokkema:
> I had to think a bit about what was going on here, since I didn't remember removing a GROUP BY statement there. But of course GROUP BY was never there, adding aggregating functions by default for all custom columns caused this effect.

Yes sorry, I should have been more clear. Indeed it wasn't removed, but it was needed now because aggregating functions were added.

> In the future we may want to change the approach by only adding aggregating functions if there's actually a GROUP BY statement defined. For now, this fix seems to do the trick.

As far as I know that is already our approach :wink:  I think we just missed it here...

PR #181, opened 2017-02-07
Fixed query errors in gene graphs page.

ifokkema:
Fixed bug; Graph variantsTypeProtein() returned a query error after application of fix for SQL mode "ONLY_FULL_GROUP_BY".
- It didn't like the "FUNCTION(MIN(column)) AS alias GROUP BY alias" structure.
- Also removed the MIN() from a "MIN(column) GROUP BY column" structure in variantsTypeDNA(). It didn't cause an error, but didn't have any effect, either.

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -22,6 +22,9 @@
    entry. This was caused by the recent changes making LOVD compatible with the
    new default "ONLY_FULL_GROUP_BY" MySQL setting.
    Closes #176: "Viewlist entry count differs from what is displayed".
+ * Fixed bug; The gene's graphs pages only showed half of the graphs due to a
+   query error introduced by the recent changes making LOVD compatible with the
+   new default "ONLY_FULL_GROUP_BY" MySQL setting.
 
 
 /**************************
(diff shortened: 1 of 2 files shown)


ifokkema:
Hi @mkroon1, I'll apply this one in the shared install already since the error is flooding our logs for a while now (and I hadn't checked for a while, almost 12K log entries :cry: ). The diff is not clear on GitHub, but I just removed the MIN() function. I hope my summary above is clear enough about the why of the error. This can go into `master` whenever you had time to look at it. Thanks!

mkroon1:
Ah, thanks. Looking back at what I did in #161, I modified `graphs.php` in some places, but did not test. I did not know how probably and I thought most of that code wasn't used atm. So we have to look out for potentially more issues in that file.

ifokkema:
> Ah, thanks. Looking back at what I did in #161, I modified graphs.php in some places, but did not test. I did not know how probably and I thought most of that code wasn't used atm. So we have to look out for potentially more issues in that file.

Ah, no worries, I checked already for more issues :) In this case, it errored because the column used in the `GROUP_BY` was also modified in the `SELECT`, but that doesn't happen elsewhere in this file, so we're good. I've also verified that all graphs work.

I can imagine you didn't see how to test this; I guess simplest is to search the project code for the function names, or the inclusion of the class file, and you'll see where the graphs are implemented. There should be no unused code, that's still in the old graphs branch that I still need to fix and implement ;)

PR #182, opened 2017-02-07
Let dev/test installations identify themselves correctly

### LINKED ISSUE #174: Let test installations identify themselves correctly
Properly showing LOVD installation counts has become completely impossible due to extremely high numbers. The busiest month, September 2016, reported as many as 4000 LOVD3 installations. Apparently test installations aren't identified (anymore?) as such.
Test installations on Travis should be marked as such. There is code in place for it, but apparently that doesn't work anymore.
Test installations ran locally should also be marked as such, so that installation/deinstallation rounds don't pollute the statistics.

mkroon1:
Overwrite default signature for development and test installations based on `$_SERVER['SERVER_ADMIN']`. Fixes #174 

@ifokkema: I was not able to test on the receiving end whether the correct signature was set. Do you know how to do this? As of now there should be test installations showing up in the statistics with the following signatures (name -- md5 hash):

    travis_CI_3.0 -- 1951dc9caea4b8c053d26612d1d7eee2
    mkroon_local_3.0 -- f85e532c2b3819323effc9d3be4baa02

### DIFF
--- a/src/install/index.php
+++ b/src/install/index.php
@@ -4,13 +4,14 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-19
- * Modified    : 2017-01-25
+ * Modified    : 2017-02-07
  * For LOVD    : 3.0-19
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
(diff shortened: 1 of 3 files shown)


ifokkema:
Thanks, Mark!

Note that I replaced `(array_)key_exists()` with `isset()` as the latter is much faster. It doesn't mean anything in this code, but if we deliberately always use `isset()`, this will save minutes of execution time in other situations.

I haven't found your signature much in the Apache log files, just once these last three weeks. Could you verify your database is set to this signature, and otherwise fix that, sending me the signature that was used before? Then I can fix it in the statistics as well. Thanks!

PR #183, opened 2017-02-13
Feat/download link

### LINKED ISSUE #78: Download link missing on gene homepage
Higher-level users can check _"allow public to download variant entries"_ on a gene entry edit page. However, on the gene's "homepage", there is no download link to be found when not logged in.

For an example on the shared DB, see: [ZNF143](http://databases.lovd.nl/shared/genes/ZNF143)

mkroon1:
New link on gene homepage to download all related data. Link text shows up as:
  * "Download all public information" for anyone below LEVEL_CURATOR, which links to `download/all/gene_public/GENE_SYMBOL` (only if `allow_download` is true for gene)
  * "Download all information" for anyone with LEVEL_CURATOR or above, which links to `download/all/gene/GENE_SYMBOL` (always shown)

New "gene_public" category download will do same thing as `download/all/gene`, except that it strips non-public variants, individuals and phenotypes. If the gene's `allow_download` is true, it's also reachable for anyone below LEVEL_CURATOR.

Notable difference from expected behavior is that screenings linked to public variants but also to non-public individuals will be found in the output.

Fixes #78 
Fixes #183

### DIFF
--- a/src/download.php
+++ b/src/download.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2012-06-10
- * Modified    : 2017-02-15
+ * Modified    : 2017-02-17
  * For LOVD    : 3.0-19
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
@@ -365,6 +365,8 @@ if (($_PE[1] == 'all' && (empty($_PE[2]) || in_array($_PE[2], array('gene', 'min
                 $aObjects['Individuals']['filters']['statusid'] = array(STATUS_MARKED, STATUS_OK);
                 $aObjects['Phenotypes']['filters']['statusid'] = array(STATUS_MARKED, STATUS_OK);
 
+                $aObjects['Individuals']['filter_other']['Screenings']['individualid'] = 'id';
+
                 // Hide non-public columns.
                 $aObjectTranslations = array(
                     'VariantOnTranscript' => 'Variants_On_Transcripts',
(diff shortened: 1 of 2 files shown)


ifokkema:
Hi! Thanks for this!

I haven't looked the code yet, but I noticed the different URL in the PR description. Do we have a use case where a curator wants to get the public download, other than just logging out and making the download?

From #78:
> > Most of export functionality already exists in `download.php`. On the gene homepage we can link to `/download/all/gene/NAME`, and modify `download.php` to check if the download flag for the gene is set and filter non-public columns and records if needed.

> It would indeed by logical to use the same URL. For the VL downloads, we can open a separate issue if we want to work on that soon, too.
>
> `download.php` needs to check authorization, and proceed how you said only if user is not authorized and setting is enabled. The check for authorization can be done [here](https://github.com/LOVDnl/LOVD3/blob/master/src/download.php#L87), but we may need a different filter than just `gene`. Sadly, the code currently can process only one filter per URL. Make the filter `gene_public` or so? The `gene` filter is defined [here](https://github.com/LOVDnl/LOVD3/blob/master/src/download.php#L254), and needs to be modified to filter variants, individuals and phenotype data for status. Filtering the columns will need to be done somewhere [here](https://github.com/LOVDnl/LOVD3/blob/master/src/download.php#L489).

Perhaps the confusion started here; I suggest to map the `gene_public` filter to the `gene` URL, when the user is not authenticated and the download is configured to be enabled; if the user is authenticated, they just get all data. Shall we do that? It seems much more logical to me to use the same URL, just like all VLs have the same URL but show something different based on your authentication.

mkroon1:
>  I suggest to map the `gene_public` filter to the `gene` URL, when the user is not authenticated and the download is configured to be enabled; if the user is authenticated, they just get all data.

Implemented in 5afe4be

src/download.php
```diff
@@ -450,6 +518,43 @@
 
 
 
+if (isset($sFilter) && $sFilter == 'gene_public') {
+    // Ugly hack to run otherwise ignored filter of individualid on screenings.
+    // FIXME: make sure this filter is part of the query to get screenings from DB.
```
ifokkema: Hi @mkroon1 , can you tell me what you mean with this FIXME? I have solved the other FIXME (see a5e0ed3) but I don't understand this one.

src/download.php
```diff
@@ -87,10 +137,20 @@
...
-        lovd_requireAuth(LEVEL_CURATOR);
+
+        if ($_AUTH['level'] >= LEVEL_CURATOR) {
+            $sFilter = 'gene';
+        } else {
+            $bAllowDownload = $_DB->query('SELECT allow_download FROM ' . TABLE_GENES . ' WHERE id = ?', array($ID))->fetchColumn();
+            if ($bAllowDownload) {
+                $sFilter = 'gene_public';
+            } else {
+                die('Error: Data for gene (' . htmlspecialchars($ID) . ') is not public and you don\'t have permission '.
```
ifokkema: I've added the `htmlspecialchars()` as a precaution, in case we ever drop the regexp on the ID. Beware of XSS attacks.

src/download.php
mkroon1: The fixme simply states to find a better solution for the "ugly hack" in this block. The code would be easier to understand if the filters in this block are handled above in a similar way to the other filters.

I don't have a solution in mind, so it serves more as a reminder than a concrete point of action.

PR #186, opened 2017-02-16
Feat/variants in range

### LINKED ISSUE #57: Make all variants within a genomic range viewable
A possible use case is that someone wants to see all variants within a certain genomic range.

At the moment I see two possible solutions to provide such functionality:
1. Show `position_g_start` and `position_g_end` columns in the variant viewlist and let users set boundaries (e.g. `position_g_start > 123`).
2. All variants of a single chromosome can already be viewed by visiting page `variants/chrX`. This can be extended to allow a finer selection with `variants/chr:start-end` (e.g. `variants/chrX:123-456`)

Personally my preference is to go for option (2), since setting boundaries via the viewlist is a bit tedious.

Depends on #58.

mkroon1:
Extending the view for all variants on a chromosome (e.g. `variants/chrX`) to make it possible to specify a subregion (e.g. `chr3:20-200,000`).

Note: the change in 4c910c9 moves the position fields settings for the viewlist into the VOG class definition. This causes the position fields to be ignored in the error message that appears when a region is specified for where there are no variants. E.g. `variants/chr21:1-2` will show a message like: `No variants found where chr is "21"!`. Code responsible for this message is [here](https://github.com/LOVDnl/LOVD3/blob/feat/variants_in_range/src/class/objects.php#L2437-L2450)

Fixes #57

### DIFF
--- a/src/class/object_genome_variants.php
+++ b/src/class/object_genome_variants.php
@@ -4,8 +4,8 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-12-20
- * Modified    : 2017-02-16
- * For LOVD    : 3.0-18
+ * Modified    : 2017-04-24
+ * For LOVD    : 3.0-19
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivar C. Lugtenburg <I.C.Lugtenburg@LUMC.nl>
@@ -175,10 +175,10 @@ class LOVD_GenomeVariant extends LOVD_Custom {
                                     'db'   => array('vog.chromosome', 'ASC', true)),
                         'position_g_start' => array(
                                     'view' => false,
-                                    'db'   => array('position_g_start', 'ASC', 'INT')),
+                                    'db'   => array('vog.position_g_start', 'ASC', true)),
                         'position_g_end' => array(
                                      'view' => false,
-                                    'db'   => array('position_g_end', 'ASC', 'INT')),
+                                    'db'   => array('vog.position_g_end', 'ASC', true)),
                       ),
                  $this->buildViewList(),
                  array(
--- a/src/class/object_genome_variants.php
+++ b/src/class/object_genome_variants.php
@@ -4,10 +4,10 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-12-20
- * Modified    : 2016-12-13
+ * Modified    : 2017-02-16
  * For LOVD    : 3.0-18
  *
- * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
+ * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivar C. Lugtenburg <I.C.Lugtenburg@LUMC.nl>
  *               Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
  *               Daan Asscheman <D.Asscheman@LUMC.nl>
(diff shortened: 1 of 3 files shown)


src/class/object_genome_variants.php
```diff
@@ -173,6 +173,12 @@ function __construct ()
                         'chromosome' => array(
                                     'view' => array('Chr', 50),
                                     'db'   => array('vog.chromosome', 'ASC', true)),
+                        'position_g_start' => array(
+                                    'view' => false,
+                                    'db'   => array('position_g_start', 'ASC', 'INT')),
+                        'position_g_end' => array(
+                                     'view' => false,
+                                    'db'   => array('position_g_end', 'ASC', 'INT')),
```
ifokkema: If you add `vog` to the column names, LOVD will know it's an INT column and the last two array elements (`'ASC', 'INT'`) won't be needed ;)

(the `ASC` is only needed if you want to make it sortable which won't work here since the column in hidden, but something needs to be specified if you need that 3rd element; so if you won't need the last anymore, you can drop the `ASC` as well)

src/variants.php
```diff
@@ -69,36 +69,46 @@
 
 
 
-if (!ACTION && (empty($_PE[1]) || preg_match('/^chr[0-9A-Z]{1,2}$/', $_PE[1]))) {
+if (!ACTION && (empty($_PE[1]) ||
+    preg_match('/^(chr[0-9A-Z]{1,2})(:([0-9,]+)-([0-9,]+))?$/', $_PE[1], $aRegionArgs))) {
```
ifokkema: Do we really want to support the `,` in the positions? It's something ugly the genome browsers are doing, but none of our APIs support it. Was this requested, or did it just seem useful? If we support it here, should our APIs also support it?

Oh, and could you add one more tab to this line, so it's clear we're still in an `if` that is not on the same level as `!ACTION`?

src/variants.php
```diff
@@ -69,36 +69,46 @@
...
+        $aColsToHide[] = 'chromosome';
+
+        if (count($aRegionArgs) == 5) {
+            // Set search conditions for start and end of region.
+            $_GET['search_position_g_start'] = '>' . str_replace(',', '', $aRegionArgs[3]);
+            $_GET['search_position_g_end'] = '<' . str_replace(',', '', $aRegionArgs[4]);
+            $sTitle .= ' in region ' . $aRegionArgs[0];
+        } else {
+            $sTitle .= ' at chromosome ' . substr($aRegionArgs[1], 3);
+        }
```
ifokkema: You could make this code a bit more readable by running a `list()` on `$aRegionArgs`, so that the values are clear. Anyone can read back of course, but it's much nicer to have a `$sChr`, `$nPositionStart`, and `$nPositionEnd` (for instance). You can then also get rid of the combined position fields by telling `preg_match()` not to count that subpattern using `?:`, like so:
```
preg_match('/^(chr[0-9A-Z]{1,2})(?::([0-9,]+)-([0-9,]+))?$/', $_PE[1], $aRegionArgs)
```
(note that then you'll need to change the check on the `count()` to 4 then)

src/class/object_genome_variants.php
mkroon1: Adding `vog.` to the database field name and removing `'INT'` breaks things. Somehow the position search conditions are ignored.

I've left this untouched for now.

src/variants.php
mkroon1: I agree. This is not the place to allow commas in the region description. If needed, that can be added in the front-end later on. Fixed in 7881a95

src/variants.php
mkroon1: Cleaned up code in 7881a95

src/class/object_genome_variants.php
ifokkema: Sorry, my bad, the 3rd argument needs to be `true` to indicate you can search on it... Fixed that in 98adab9.

PR #188, opened 2017-02-17
Better error text when no transcripts with underlying variants are found

### LINKED ISSUE #89: Misleading text "No transcripts or variants found"
For genes with a transcript, but no underlying variants, clicking the _Variants_ tab shows an info box with the text "No transcripts or variants found for this gene.".

Example on shared database: [A1CF](http://databases.lovd.nl/shared/variants/A1CF)

mkroon1:
Fixes #89


ifokkema:
How about we add a count on the transcripts in the query that checks for the gene's presence? Then we'll actually be able to really improve the text (without adding a query), and clearly indicate whether there are no transcripts, or no variants...

PR #198, opened 2017-03-03
Simplify query to check DBID

### LINKED ISSUE #185: SQL only full group by error in DBID check.
Currently the selenium import tests fail on mysql 5.7.17 with default config while running a query to check a DBID. This looks to be a remnant of issue #161.

The faulty query is generated [here](https://github.com/LOVDnl/LOVD3/blob/b200c8bc53fdab28dbf85079f1705eeffa99ef69/src/inc-lib-form.php#L74-L103) and looks like this:
```sql
SELECT DISTINCT
  t.geneid,
  CONCAT(IFNULL(vog.`VariantOnGenome/DNA`, ""), ";", IFNULL(GROUP_CONCAT(vot.`VariantOnTranscript/DNA` SEPARATOR ";"), "")) AS variants,
  vog.`VariantOnGenome/DBID`
FROM lovd_v3_variants AS vog
  LEFT OUTER JOIN lovd_v3_variants_on_transcripts AS vot ON (vog.id = vot.id)
  LEFT OUTER JOIN lovd_v3_transcripts AS t ON (vot.transcriptid = t.id)
WHERE
  ((REPLACE(REPLACE(REPLACE(vog.`VariantOnGenome/DNA`, "(", ""), ")", ""), "?", "") = 'g.40702876G>T' AND vog.chromosome = '15')) AND
  BINARY vog.`VariantOnGenome/DBID` = 'IVD_000001' AND vog.id != '0000000002'
GROUP BY vog.id;
```

@ifokkema: as I understand it, the query is only used to check if there's one or more variants different from the current one (vog.id = '0000000002') that have the same DBID. So, can't we simplify it with something like what's below?

```sql
SELECT COUNT(*)
FROM lovd_v3_variants AS vog
  LEFT OUTER JOIN lovd_v3_variants_on_transcripts AS vot ON (vog.id = vot.id)
  LEFT OUTER JOIN lovd_v3_transcripts AS t ON (vot.transcriptid = t.id)
WHERE
  vog.`VariantOnGenome/DBID` = 'IVD_000001' AND vog.id != '0000000002';
```

mkroon1:
Simplify query to check DBID for mysql 5.7.5+ default full group by setting.

Fixes #185

### DIFF
--- a/src/inc-lib-form.php
+++ b/src/inc-lib-form.php
@@ -4,10 +4,10 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-21
- * Modified    : 2016-10-14
- * For LOVD    : 3.0-18
+ * Modified    : 2017-03-02
+ * For LOVD    : 3.0-19
  *
- * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
+ * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ing. Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
  *               Ing. Ivar C. Lugtenburg <I.C.Lugtenburg@LUMC.nl>
  *               M. Kroon <m.kroon@lumc.nl>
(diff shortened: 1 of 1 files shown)


src/inc-lib-form.php
```diff
@@ -71,11 +71,8 @@ function lovd_checkDBID ($aData)
     $nHasDBID = $_DB->query('SELECT COUNT(id) FROM ' . TABLE_VARIANTS . ' WHERE `VariantOnGenome/DBID` = ? AND id != ?', array($aData['VariantOnGenome/DBID'], $nIDtoIgnore))->fetchColumn();
     if ($nHasDBID && (!empty($sGenomeVariant) || !empty($aTranscriptVariants))) {
         // This is the standard query that will be used to determine if the DBID given is correct.
-        $sSQL = 'SELECT DISTINCT t.geneid, ' .
-                'CONCAT(IFNULL(vog.`VariantOnGenome/DNA`, ""), ";", IFNULL(GROUP_CONCAT(vot.`VariantOnTranscript/DNA` SEPARATOR ";"), "")) as variants, ' .
-                'vog.`VariantOnGenome/DBID` ' .
+        $sSQL = 'SELECT count(*) ' .
```
ifokkema: Be careful :wink:
Replacing a normal select with a count(*) will change the way the results are returned, so then also handle that change in the fetching of the results. This change introduced a bug, where any variant can be assigned any given DBID of any other variant. Fixed in 665d952.

PR #227, opened 2017-06-01
Let viewlist rows open link in new window with middle-click.

### LINKED ISSUE #196: Rows in viewlist can't be opened in new tab.
Most browsers allow to open links in a new tab or window by middle-clicking a link or by right-clicking a link and choosing "open link in new tab" or something similar. The rows in most (all?) viewlists are clickable, but redirection is done via javascript, and so opening the target in a new tab or window in the manner described above is not possible.

mkroon1:
Let users open a link from a row in a viewlist by clicking with the middle button of the mouse. Pure javascript row links are unaffected.

Fixes #196.

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -46,6 +46,9 @@
      contained more text than just an GenBank record ID.
  * Updated all links to the HGVS nomenclature to point to the new website.
  * Updated all links to the NCBI to use HTTPS.
+ * Added feature to open links in tables in a new browser tab. Does not work
+   for some older browsers based on webkit.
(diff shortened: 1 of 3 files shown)


ifokkema:
This doesn't really work for me on Chromium 37.0.2062.120. It does open the new window, but it doesn't cancel the default `onclick()` behavior, resulting in the current window changing to the clicked link as well. It does work on Firefox 45.3.0. Could you test on your Chrome/Chromium installation?
I tried the following, to no avail:
```javascript
e.stopPropagation();
e.preventDefault();
e.cancelBubble=true;
return false;
```

PR #241, opened 2017-07-05
Fix problem with converting submitter IDs in the LOVD2 to LOVD3 converter.

### LINKED ISSUE #238: Submitter IDs not translated in LOVD2 conversion
The LOVD2 export conversion script (`scripts/convert_lovd2.php`) does not properly translate the submitter IDs according to the translation table given by the user in the web form.

I deduced the bug was introduced in 5d83fd7.

ifokkema:
Fix problem with converting submitter IDs in the LOVD2 to LOVD3 converter.
- Fixes #238.


ifokkema:
@mkroon1 , I noticed that this test will fail on any new LOVD release, since the LOVD3 version is contained in its output file. So we either need to renew the expected LOVD3 output file every single release (meh), or update the code to ignore the version (or just the first line or so, I wouldn't mind). What do you think?

Also, could you add a changelog entry?

Issue #216, opened 2017-05-11
HGVS notation for mitochondrial positions.


mkroon1:
I presume the following issue also affects LOVD:

mutalyzer/mutalyzer#425

I have seen code within LOVD that especially accounts for the `m.` notation for mitochondria. The full extent of changes needed to completely convert to `g.` notation still has to be investigated.

mkroon1:
Note that data in `VariantOnGenome/DNA` will have to be translated from "m.*" to "g.*".

mkroon1:
This issue is currently on hold as the change to the [guidelines](http://varnomen.hgvs.org/bg-material/refseq/#DNAg) to drop the `m.` notation is still a matter of debate. The fix in branch `fix/hgvs_mitochondrial` will remain unmerged until the change is accepted.

Issue #58, opened 2016-04-19
Extract positions from variant description


mkroon1:
Variant descriptions are parsed and the genomic positions (start, end) are stored separately in the database. However, if parsing fails (via Mutalyzer) the genomic positions are not stored.

LOVD could try to extract the positions from the description heuristically. This would benefit for example the feature described in #57 
- [x] Download all variants from shared LOVD without genomic positions, and test if we can extract those positions.
- [x] Verify with Johan which positions should be taken in case of unclear boundaries.
- [x] Use our method instead of Mutalyzer, whenever we can.

PR #258, opened 2017-07-24
Added script that analyses variant position problems and will fix them, and improved the lovd_getVariantInfo() function.

ifokkema:
Added script that analyses variant position problems and will fix them, and improved the lovd_getVariantInfo() function.
- The script analyses some of the possible errors, and reports them.
  - Swapped position fields.
  - Incorrect variant positions.
  - Missing variant positions.
  - Variants that are not understood by the normal process (=Mutalyzer).
- When you press run, it will fetch the problematic data, run a function over the data, and updates the table and progress bar.
- Updated styles.css to allow simple highlighting of a row in an data table.
- Fixed warning when Mutalyzer returns false on getting a gene's location for generating an UD.
- Updated lovd_getVariantInfo() to return the position fields correctly when the DNA field has them in the wrong order.
  - Note that this function does not have the power to fix the DNA field, it simply stores the positions the way they should have been.
- Added support for variants with uncertain positions in lovd_getVariantInfo().
  - The positions returned are calculated as the average of the uncertain positions, instead of taking the minimum (shortest, certain) positions of the variant.
  - May need tweaking or additions later.
- Fixed bug; lovd_getVariantInfo() would sometimes miscall the position_c_end_intron field.
- Improved lovd_getVariantInfo() to include maximum and minimum values to the position fields.
  - This prevents query errors when updating the position fields, in case of a strict SQL mode.
  - Also, made sure the c. and n. variants always have the intronic fields defined for variants with uncertain positions, just like variants with certain positions.

Related: #58.

After review, still to do:
- Fully implement #58; use this function for all variant position calculations.


mkroon1:
Running the current version of the `fix_position_fields.php` script resulted in a query error during the `vot_positions_in_error` analysis:

Query : UPDATE lovd_v3_variants_on_transcripts SET position_c_start = ?, position_c_start_intron = ?, position_c_end = ?, position_c_end_intron = ? WHERE id = ? AND transcriptid = ?
Error : Error in PDOStatement::execute() while executing prepared query: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'position_c_start' at row 1

The causal query:

    UPDATE lovd_v3_variants_on_transcripts SET position_c_start = '191861921', position_c_start_intron = '0', position_c_end = '191861921', position_c_end_intron = '0' WHERE id = '0000039881' AND transcriptid = '00006722'

That particular variant description is ridiculous: `c.191861921delTCAG`

Other than that I don't see any blocking issues.

ifokkema:
> That particular variant description is ridiculous: c.191861921delTCAG

I think it's fair enough to say this variant is impossible; there's only three chromosomes in which this even fits, and if it does, the variant will be all the way on the other side of the chromosome. If we want to prevent inserting these, I guess we'd need to fix that elsewhere.

If you agree on not fixing this, we can merge this?

mkroon1:
I agree that issue may be fixed somewhere else. But it would be nice if the script didn't block on the error and continue to fix as much as possible.

ifokkema:
> I agree that issue may be fixed somewhere else. But it would be nice if the script didn't block on the error and continue to fix as much as possible.

The simplest would be to simply ignore any query errors? Could be accomplished by adding falses to the query calls, although to prevent a "method on non-object" error, we'll need to first run the query, then check the result, and then do the `rowCount()` and let the functions return the array. Would that be OK for you?

mkroon1:
Yeah, I think that would be sufficient. I think the update queries are the most important, as far as I can see those all return a row count and not an actual object. Adding false to those would be enough imo.

src/inc-lib-init.php
```diff
@@ -567,20 +567,114 @@ function lovd_getVariantInfo ($sVariant, $sTranscriptID = '')
...
+        //                                                                                                                  8 = Latest end position, might be a question mark.
+        //                                                                                                                          9 = Latest end position intronic offset, if available.
+        //                                                                                                                                              10 = The variant, which we'll use to determine the type.
+
+        list(, $sPrefix, $sStartPositionEarly, $sStartPositionEarlyIntron, $sStartPositionLate, $sStartPositionLateIntron, $sEndPositionEarly, $sEndPositionEarlyIntron, $sEndPositionLate, $sEndPositionLateIntron, $sVariant) = $aRegs;
+
+        // Store positions.
+        // If each position (start, end) has two numeric positions, we choose the
+        //  middle one (latest start, earliest end). Otherwise, we pick the numeric one.
+        // We do require at least one numeric start position and one numeric end position.
```
ifokkema: Johan has requested to take the average of the two numeric positions, and not the middle ones.

src/changelog.txt
```diff
@@ -29,6 +29,8 @@
    visitors to the homepage will be redirected to that gene's page instead of
    the gene list.
    Closes #98: "Configurable homepage".
+ * Added script to fix known errors in variant positions. The script has been
+   added to the scripts folder.
```
ifokkema: Johan has requested to fully implement #58. Therefore, make sure that this function is implemented everywhere, and add this to the changelog.

ifokkema:
> Running the current version of the fix_position_fields.php script resulted in a query error during the vot_positions_in_error analysis

Fixed in 7abac4a80ad37a9fc2df906f7e7062e843d4c14e, by letting `lovd_getVariantInfo()` return values that are maximized to the upper and lower limits of the fields.

mkroon1:
Re-running the `fix_position_fields.php` script on a copy of the shared now causes about two dozen instances of the notice below. It appears that `lovd_getVariantInfo()` sometimes returns an array without key `position_start_intron`.


    Notice: Undefined index: position_start_intron in /home/mkroon/LOVD/LOVD3_dev/src/scripts/fix_position_fields.php on line 190
    --
    1 | 0.0027 | 360384 | {main}( ) | .../fix_position_fields.php:0
    2 | 0.1921 | 480432 | LOVD_VariantPositionAnalyses->runAnalyses( ) | .../fix_position_fields.php:439
    3 | 41.9636 | 55035488 | LOVD_VariantPositionAnalyses->{closure:/home/mkroon/LOVD/LOVD3_dev/src/scripts/fix_position_fields.php:178-196}( ??? ) | .../fix_position_fields.php:312

mkroon1:
Commit b39b8fb removes support for mitochondrial "m.\*" descriptions, intended to solve issue #216

ifokkema:
> Re-running the fix_position_fields.php script on a copy of the shared now causes about two dozen instances of the notice below. It appears that lovd_getVariantInfo() sometimes returns an array without key position_start_intron.

Are you sure you're using the latest code? That should already have been fixed. Only for genomic positions it won't generate those fields.

> Commit b39b8fb removes support for mitochondrial "m.*" descriptions, intended to solve issue #216

We can not do that unless we're sure that no LOVD installations use `m.` anymore. We need to be backwards compatible.

ifokkema:
> The problem is that somebody messed up and entered a "g.\*" description in the `VariantOnTranscript/DNA` field -- for which `getVariantInfo()` obviously does not return intron positions. I suppose fixing such things automatically is impossible and thus out of scope for this script. So, no change needed.

Not impossible - we could add a check with fixing function in the analysis script to fix the variant's prefixes as well, if we want to.... That would fix these in our LOVD, and whomever would get the script and runs it as well. We would however need the transcript info to see whether a `g.` needs to be a `c.` or an `n.` variant.

> Do you propose not to fix #216 at all? (e.g. changes in branch [fix/hgvs_mitochondrial](https://github.com/LOVDnl/LOVD3/tree/fix/hgvs_mitochondrial) hide the "map to transcript" button when a "m.\*" description is specified for the DNA field.)

I propose to push people towards the new notation, without breaking existing implementations. We could disallow people to use `m.`, although instead of nagging them with removed buttons and broken services, we'd better making the change automatically or clearly warn them about the changed nomenclature so they can make the change themselves :wink:

mkroon1:
>> The problem is that somebody messed up and entered a "g.*" description in the VariantOnTranscript/DNA field -- for which getVariantInfo() obviously does not return intron positions. I suppose fixing such things automatically is impossible and thus out of scope for this script. So, no change needed.

> Not impossible - we could add a check with fixing function in the analysis script to fix the variant's prefixes as well, if we want to.... That would fix these in our LOVD, and whomever would get the script and runs it as well. We would however need the transcript info to see whether a g. needs to be a c. or an n. variant.

Finding a "g.\*" description on the transcript level usually mean 2 mistakes I believe: 1) the prefix is wrong as it should be "c.", or 2) the description is placed in the wrong field and should be in `VariantOnGenome/DNA`. Assuming it's always mistake 1 is easy, I suppose.

Other than that, I don't think there's anything to be done for this PR, and it can be merged.

src/changelog.txt
ifokkema: Will do this later, separately from this branch.

PR #265, opened 2017-08-04
Added links to overviews of created entries on user page.

### LINKED ISSUE #190: Add links to records created by user on user's profile page
on the "user account profile" a user can see the number of records Owned, but not the number of records Created. Please add this.
I just was in contact with a user that could not find records originally created but that were moved to another Owner when she changed working places. Regarding"micro-attribution", showing who contributed what, this would be a nice addition.

mkroon1:
* Links to overviews are similar to those for data owned by the given
  user.
* Edited individual and screening classes to include created_by in
  their viewlist (hidden but searchable).

Fixes #190

### DIFF
--- a/src/class/object_users.php
+++ b/src/class/object_users.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-21
- * Modified    : 2017-08-04
+ * Modified    : 2017-08-10
  * For LOVD    : 3.0-20
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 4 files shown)


src/class/object_users.php
```diff
@@ -516,10 +522,28 @@ function prepareData ($zData = '', $sView = 'list')
...
+                    if ($n) {
+                        $nCreates += $n;
+                        $sTitle = $n . ' ' . ($n == 1? substr($sDataType, 0, -1) : $sDataType);
+                        // Hide the link for phenotypes, because we don't have a phenotypes
+                        // overview to link to (must be disease-specific).
+                        $sCreates .= (!$sCreates? '' : ', ') . ($sDataType == 'phenotypes'?
+                                $sTitle : '<A href="' . $sDataType . '?search_created_by=' .
+                                rawurlencode(html_entity_decode($zData['id'])) . '">' .
+                                $sTitle . '</A>');
+                    }
```
ifokkema: Perhaps it would be worth merging these two code blocks a bit more? You could save 4 queries to the database, and by adding another loop like `foreach (array('owned_by', 'created_by') as $sField)` you can reduce the code duplication even more. Do you want to go for it?

PR #267, opened 2017-08-07
Clearer description of panel_size field for individuals.

### LINKED ISSUE #191: Panel size description could be improved
on the submission form, Individual, the text linked to the **Panel size** could be improved. I suggest:

Number of individuals this entry represents; e.g. 1 for an individual, 5 for a family with 5 affecteds. To report different Individuals from one family link them using the field "ID of panel this entry belongs to".

mkroon1:
New legend text for panel size. Decided on following text for form:

> Number of individuals this entry represents; e.g. 1 for an individual, 5 for a family with 5 affected members. To report different Individuals from one family link them using the field "ID of panel this entry belongs to".

And for viewlist:

> Number of individuals this entry represents; e.g. 1 for an individual, 5 for a family with 5 affected members.

Fixes #191

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -31,6 +31,8 @@
    Closes #98: "Configurable homepage".
  * Fixed bug; The variant mapper gets stuck when encountering variants with no
    position set.
+ * The "Panel size" field for individuals has a clearer description.
+   Closes: #191: "Panel size description could be improved".
 
 
 /**************************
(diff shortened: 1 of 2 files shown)


ifokkema:
If we're going to add this:

> To report different Individuals from one family link them using the field "ID of panel this entry belongs to".

Then I think we need to be clearer on what we mean here. I wouldn't be able to explain that in one sentence, so I'm guessing this short description will confuse users.

PR #268, opened 2017-08-08
Fix effect_reported not set by curator

### LINKED ISSUE #213: Curator is allowed to leave `effect_reported` set to unclassified.
~`effect_reported` should not be "unclassified" in any circumstance.~

Do not allow `effect_reported` to be set to "not classified" when statusid is "Public".

mkroon1:
Do not allow `effect_reported` to be set to "not classified" when statusid is "Public".

Fixes #213

### DIFF
--- a/src/class/object_genome_variants.php
+++ b/src/class/object_genome_variants.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-12-20
- * Modified    : 2017-08-08
+ * Modified    : 2017-08-10
  * For LOVD    : 3.0-20
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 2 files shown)


src/class/object_genome_variants.php
```diff
@@ -245,9 +245,19 @@ function checkFields ($aData, $zData = false)
...
+            // `effect_reported` is not allowed to be '0' (Not classified) when user is a submitter
+            // or when the variant has status '9' (Public).
+            if ($_AUTH['level'] < LEVEL_CURATOR) {
+                // Remove the mandatory `effect_reported` field to throw an error.
+                unset($aData['effect_reported']);
+            } elseif (isset($aData['statusid']) && intval($aData['statusid']) == 9) {
+                // Show error for curator/manager trying to publish variant without effect.
+                lovd_errorAdd('effect_reported', 'The \'Affects function (reported)\' field ' .
+                    'must be filled when variant status is Public');
+            }
```
ifokkema: - To prevent problems later, please use `STATUS_PUBLIC` instead of `9`.
  We could decide on applying this method as well to the variant effects?
- We don't need (or use) `intval()` anywhere. Would there be an advantage? We always rely on the typecasting of the `==` operator.
- I think the error message could be improved, as it's pretty much the same as the user will get when the field is really not filled in. Perhaps add that `Not classified` is not allowed for public variants?

mkroon1: Fixed with 0a8e1eb according to your suggestions: now using `STATUS_OK` and adjusted error message.

PR #278, opened 2017-08-17
Reposition content area based on header size even when no announcements.

### LINKED ISSUE #277: The page header isn't properly displayed when system logo is changed.
The page header isn't properly displayed when system logo is changed, if the image is bigger than the original logo we cant access the features links.
e.g.:http://bipmed.iqm.unicamp.br/LOVD/genes

mkroon1:
The content area in HTML is now repositioned using javascript based on
the actual height of the sticky header. The predicted height may be
wrong even when there are zero announcements. (e.g. when the standard
logo is replaced with a larger image like in issue #277).

Fixes #277

### DIFF
--- a/src/class/template.php
+++ b/src/class/template.php
@@ -4,8 +4,8 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2012-03-27
- * Modified    : 2017-01-25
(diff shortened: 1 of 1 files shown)


ifokkema:
Feel free to merge whenever you want, I'm getting:

> Pull requests that have a failing status can’t be merged on a phone.

PR #282, opened 2017-09-01
Remove lingering phenotype objects set during import.

### LINKED ISSUE #245: Import file parsing fails with a fatal error when the diseaseid field is left empty for a phenotype record
Import file parsing fails with a fatal error when the diseaseid field is left empty for a phenotype record.

`Fatal error: Call to a member function getForm() on null in import.php on line 850.`

getForm() should not be run if we don't have an disease ID.

mkroon1:
* Fix #245 by making sure that a phenotype object set during parsing of
  a previous line is removed when field `diseaseid` is empty, in order
  to avoid that the wrong phenotype object is used in later processing
  of the current input line.

Fixes #245

### DIFF
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -44,6 +44,10 @@
    unclassified."
  * Removed redundant "View" in all page titles. E.g. "View all genes" became
    "All genes".
+ * Fixed bug; Importing a phenotype entry without specifiying a disease ID
+   resulted in a fatal error.
+   Closes #245: "Import file parsing fails with a fatal error when the diseaseid
+   field is left empty for a phenotype record".
 
 
 /**************************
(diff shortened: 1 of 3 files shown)


ifokkema:
Hi @mkroon1 , I've pushed some additional changes to make sure the VOTs were checked equally to the phenotype entries now. Could you check my changes, and merge if you're OK with them?

mkroon1:
A test import file has a VOT record with a non-existent transcriptid. Previously the test ran checkFields() on the object from the previous record (with whatever gene was linked there), the test would then check for the error thrown by checkFields(). Fixed this in 50ec349

However, when checkFields() cannot be run, as in this case, the import script does further checks (e.g. is the pos_start field smaller than pos_end?). These checks may be futile without having run checkfields() (e.g. why check if pos_start is smaller when it's not even an integer?). Maybe it's better to not do any additional checks at all when checkFields cannot be called.

PR #287, opened 2017-09-20
Feat/additional variant effects

### LINKED ISSUE #261: Additional variant effect values
There is need for additional variant effect values. Currently, when a variant does have an effect on the function of the gene, but may not cause disease or may not cause the disease associated with the individual in which the variant is reported, it is debatable which variant effect should be selected. Therefore, it is proposed to add two additional values.

The currently active effects:

| Code | ID | Description |
|---|:-:|---|
| . | 0 | Not classified |
| ? | 5 | Variant effect unknown |
| + | 9 | Variant affects function |
| +? | 7 | Variant probably affects function |
| -? | 3 | Variant probably does not affect function |
| - | 1 | Variant does not affect function |

It is proposed to add:

| Code | ID | Description |
|---|:-:|---|
| # | 6 | Variant affects function but was not associated with any known disease phenotype |
| +* | 8 | Variant affects function but was not associated with this individual's disease phenotype |

To be determined:
- Perhaps the description of # should read "disease" instead of "disease phenotype"?
- What to do with these values in the API?
  - Translate `#` to `-` to indicate it's not disease-causing, or add as separate value that might not fit in other system's 5-class lists?
  - Translate `+*` to `+` to indicate it's disease causing, or add as separate value that might not fit in other system's 5-class lists?

mkroon1:
New variant effect IDs "#" and "+\*". 

* Added new codes and descriptions to $_SETT['var_effect']. 
* Updated legend in 3 viewlists. 
* Modified submitter test to use new effect ID. 
* New way of generating reported/concluded effect symbol combinations. 
* Added new effect symbol combinations to SQL upgrade.

Fixes #261

### DIFF
--- a/src/2Do.txt
+++ b/src/2Do.txt
@@ -22,8 +22,12 @@ To debug: XDEBUG_PROFILE=1
 
 Prioritize:
 ===========
+Columns we no longer use that can be removed:
+- TABLE_GENES/allow_index_wiki
+- TABLE_TRANSCRIPTS/id_mutalyzer
+- TABLE_USERS/reference
+
 \item[$\circ$] LSDB list: Adding data from dead LSDBs: Brenda?
-\item[$\circ$] Really implement the deletion range recognition. Store inner positions, if 4 given.
 
 Website:
 - Create search interface w/ Ajax for ww query service, but in a secure way, so it can't be abused.
--- a/src/changelog.txt
+++ b/src/changelog.txt
@@ -55,7 +55,7 @@
    Closes #242: "Display of screening with linked variants".
  * New effect categories for variants that affect protein function but are
    either not associated with the individual's disease phenotype or not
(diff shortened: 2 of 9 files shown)


ifokkema:
Going through this, I'm getting confused. I was sure the new value 6 was used for "Variant affects function but was not associated with any known **disease** phenotype" (as in the issue), but the implementation is "Variant affects function but was not associated with any known phenotype".

These are two completely different things, and we need to sort out the requirements before we can proceed. I don't see the advantage to implement it the way it's done, but we need clarification from what Johan wanted.

mkroon1:
Reworded this in a2c032f to:

"Affects function, not associated with any known disease phenotype"

ifokkema:
I added "disease" in the other places where it had been left out as well, and added the new variant effects to the legend in two more places. I'll merge the branch now.

PR #288, opened 2017-09-22
Fixed the variant position calculation

### LINKED ISSUE #58: Extract positions from variant description

ifokkema:
Fixed the variant position calculation:
- Replaced all calls to Mutalyzer's mappingInfo that were not also mapping requests by calls to the lovd_getVariantInfo() function.
- Fixed bug; When automatically mapping variants of genes on the antisense strand, the variant positions were stored in the wrong order.

Closes #58.

### DIFF
--- a/src/ajax/map_variants.php
+++ b/src/ajax/map_variants.php
@@ -122,6 +122,8 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
 
         // Loop the transcripts and map the variant to them.
         foreach ($aTranscripts as $aTranscript) {
+            // FIXME: When manually mapping transcripts, the numberConversion is just called once, and lovd_getVariantInfo() subsequently for the position fields.
+            // That's probably more efficient than mappingInfo a bunch of times.
             if (empty($aTranscript['id_ncbi'])) {
                 // A transcript without accession number is encountered. Invalid arguments, return false.
                 return false;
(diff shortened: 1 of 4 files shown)


src/ajax/map_variants.php
```diff
@@ -147,6 +149,13 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
             foreach ($aVariantsOnTranscripts[$sVariant] as $sVariantOnTranscript) {
                 if (substr($sVariantOnTranscript, 0, strlen($aTranscript['id_ncbi'])) == $aTranscript['id_ncbi']) {
                     // Got the variant description relative to this transcript.
+                    // 2017-09-22; 3.0-20; The mappingInfo module call does not sort the positions, and as such the "start" and "end" can be in the "wrong" order.
```
mkroon1: What is the reason for these type of comments:

> // 2017-09-22; 3.0-20; ...

They show removed code, but otherwise they seem superfluous given that everything is available in git.

src/variants.php
```diff
@@ -699,18 +694,20 @@ function(sData) {
...
+                        if ($aResponse) {
+                            $_POST[$nTranscriptID . '_position_c_start'] = $aResponse['position_start'];
+                            $_POST[$nTranscriptID . '_position_c_start_intron'] = $aResponse['position_start_intron'];
+                            $_POST[$nTranscriptID . '_position_c_end'] = $aResponse['position_end'];
+                            $_POST[$nTranscriptID . '_position_c_end_intron'] = $aResponse['position_end_intron'];
+                        } else {
+                            $_POST[$nTranscriptID . '_position_c_start'] = 0;
+                            $_POST[$nTranscriptID . '_position_c_start_intron'] = 0;
+                            $_POST[$nTranscriptID . '_position_c_end'] = 0;
+                            $_POST[$nTranscriptID . '_position_c_end_intron'] = 0;
```
mkroon1: My gut feeling is that it's more clear to store unkown positions as NULL instead of 0.

src/ajax/map_variants.php
ifokkema: What do you mean "removed code"? The date and version number mentioned is indeed not really needed anymore since we have everything in git (it's merely a quick lookup/reminder of when the bugfix took place), but the comment itself is still valuable as it explains what is done below.

src/variants.php
ifokkema: No, `NULL` is for variants where the position fields haven't been filled in.... so I need something else to indicate "we've looked at it, but I can't identify it". 0 seems the most logical choice...

src/ajax/map_variants.php
mkroon1: Ah I linked this comment to the wrong line, sorry about that. Line 675:

> // 2017-09-22; 3.0-20; Replacing the old SOAP call to Mutalyzer with our new lovd_getVariantInfo()

This mentions removed code (old SOAP call).

src/variants.php
mkroon1: Shouldn't lines 680-681 then also be changed?

                 $_POST['position_g_start'] = NULL;
                 $_POST['position_g_end'] = NULL

ifokkema: Yes, you're right, and also in 2449-2451. In some places this was used assuming the SOAP call had simply failed, and we should try again at some other time. Now we'll get a false only when the data isn't correct, so it should be changed. I'll take care of it, thanks for picking it out!

ifokkema: Done in b0cdf0b.

PR #294, opened 2017-09-29
Filter viewlists based on multiple values for a column

### LINKED ISSUE #202: Overview of variants with multiple classifications
The view `variants/GENE/unique` shows variants with possibly multiple classifications (column "effect") grouped together. Of these, the ones with multiple (and possibly contradictory) classifications are of most interest, as they depict variants that may require attention.

However, it is not possible to get an overview of the variants with multiple classifications. For instance, [searching for `,` in the "effect" column in the view mentioned above](http://databases.lovd.nl/shared/variants/ABCA4/unique#object_id=VariantOnTranscriptUnique%2CVariantOnGenome&id=ABCA4&order=VariantOnTranscript%2FDNA%2CASC&skip[chromosome]=chromosome&skip[allele_]=allele_&skip[transcriptid]=transcriptid&search_transcriptid=00001847&skip[owner_countryid]=owner_countryid&FRMatchType_CustomVL_VOTunique_VOG_ABCA4=1&FRMatchType_CustomVL_VOTunique_VOG_ABCA4=2&FRMatchType_CustomVL_VOTunique_VOG_ABCA4=3&FRReplaceAll_CustomVL_VOTunique_VOG_ABCA4=1&=Preview&=Cancel&=Submit&search_vot_effect=%2C&page_size=100&page=1) to see only the rows with multiple values for that column does not yield the desired result.

mkroon1:
Note: currently only activated in the unique variants view.

Fixes #202

### DIFF
--- a/src/class/objects.php
+++ b/src/class/objects.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-21
- * Modified    : 2017-09-29
+ * Modified    : 2017-10-06
  * For LOVD    : 3.0-20
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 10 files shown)


ifokkema:
I've added some commits, mainly to make `viewList()` not have tons of arguments. I now created an `$aOptions` array argument, containing the find and replace flag, and the new multiple value filter flag. Later we can add more, but now I just want to get this merged :wink:

Test succeeded, so merging!

PR #299, opened 2017-10-09
Add variant effect to the LOVD2-style API

ifokkema:
The LOVD2-style API can now optionally release the variant effect fields.
- The LOVD2-style API can now release the variant effect fields, when asked for them. They are requested through the show_variant_effect flag.
- It is done this way to prevent the default output to change, possibly breaking clients of the API.
- Brought over lovd_mapCodeToDescription() from LOVD+ to LOVD3.
- Values are human readable, but more concise than the values used within the LOVD3 interface itself.
- Values are separated by reported and concluded, and multiple values (for multiple observations) are separated by comma.


ifokkema:
@mkroon1 I would normally ask for a code review, but I'm afraid I'll skip that today due to the need to get a release out :wink: 

Feel free to browse through the code anyway!

PR #303, opened 2017-10-17
Fix issue with default value for parameter in javascript function.

mkroon1:
* Some of the current browsers (e.g. IE11) don't accept default argument values in JS.


ifokkema:
This PR didn't show the other commits made in this branch for some reason, closing it.

PR #304, opened 2017-10-17
Fix broken viewlists in Internet Explorer.

ifokkema:
Recent changes have introduced a default argument in one of the new JS functions in `inc-js-viewlist.php`. However, Internet Explorer does not support this, and this breaks all VLs.


ifokkema:
Merged in master, somehow GitHub doesn't get that.

PR #309, opened 2017-11-06
Fix submission API issues

ifokkema:
Fix submission API issues.
- Display the admin's email address better when it's multiple addresses.
- Added LSDB ID to Admin's Auth Token dialog.
  - This way the Admin can find out which ID the clients need to use to start submitting data into the LOVD.
- Fixed bug; 3'UTR variants were not accepted.
  - lovd_getVariantInfo() should be run passing the transcript ID in such cases.
- Fixed query errors coming from LOVD_Object::getRowCountForViewList().
  - This function tries to calculate the row count for each VL query, without having to use MySQL's SQL_CALC_FOUND_ROWS().
  - It takes the VL query and simplifies it as much as possible, removing unused selects, joins and sorts, and then getting the COUNT(*).
  - Searches cause query errors in this function where newlines are placed in the query definition, and aliases with capitals were not supported. Updated the query matcher so it handles this well.

### DIFF
--- a/src/class/api.submissions.php
+++ b/src/class/api.submissions.php
@@ -4,10 +4,10 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2016-11-22
- * Modified    : 2016-12-07
- * For LOVD    : 3.0-18
+ * Modified    : 2017-11-06
+ * For LOVD    : 3.0-21
  *
- * Copyright   : 2004-2016 Leiden University Medical Center; http://www.LUMC.nl/
+ * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmer  : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
  *
  *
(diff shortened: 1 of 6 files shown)


ifokkema:
Hi @mkroon1 

I needed this fix because the API didn't support submission of 3'UTR variants and our tester was running into that problem. I included other small fixes of things that I encountered. It's not much though. I would appreciate if you could take quick look so we can merge this. I already copied the fix into the shared, so there is no rush.

src/class/api.submissions.php
```diff
@@ -503,7 +503,8 @@ public function processPOST ()
         if (!$_INI['paths']['data_files']) {
             // There's no way we can do anything with the results, so fail.
             $this->API->aResponse['errors'][] = 'No data file path configured. Without this path, this LOVD API is not able to process files. ' .
-                'Please contact the admin to configure the data file path: ' . $_SETT['admin']['name'] . ' <' . $_SETT['admin']['email'] . '>.';
+                'Please contact the admin to configure the data file path: ' . $_SETT['admin']['name'] . ' <' .
+                str_replace(array("\r\n", "\r", "\n"), '>, <', trim($_SETT['admin']['email'])) . '>.';
```
mkroon1: This string replacement:
~~~php
str_replace(array("\r\n", "\r", "\n"), '>, <', trim($_SETT['admin']['email'])) . '>.'
~~~
is done 4 times in the same (rather big) function. I think it's better to put it in a variable somewhere at the top to make it more maintainable.

ifokkema: Makes sense... Implemented in 7198a5b.

PR #319, opened 2017-11-30
Fixed bug; The color coding was visible for the public.

ifokkema:
Fixed bug; The color coding was visible for the public.
- Color coding based on the variant's status should not be shown to the public, also because they don't see the status column.
- Made sure the color coding was done only for users >= LEVEL_COLLABORATOR.
- Also made sure the color coding logic was in one place only, instead of copied in many different files.
- Color coding is based on the lowest set status (unique variant view and view with variant and patient data in one row).

### DIFF
--- a/src/class/objects.php
+++ b/src/class/objects.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-21
- * Modified    : 2017-11-20
+ * Modified    : 2017-11-30
  * For LOVD    : 3.0-21
  *
  * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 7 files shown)


src/class/objects.php
```diff
@@ -1289,7 +1289,31 @@ function prepareData ($zData = '', $sView = 'list')
...
+                $zData['ind_statusid'] = ($zData['ind_statusid']?: STATUS_OK);
+                // In the VariantOnTranscriptUnique view the var_statusid can contain multiple IDs, these IDs are separated by a ",".
+                // PHP always takes the first integer-like part of a string when a string and an integer are compared.
+                // But to avoid problems in the future, only the first character is compared.
+                $nStatus = min($zData['statusid'], substr($zData['var_statusid'], 0, 1), $zData['ind_statusid']);
+                $zData['class_name'] = '';
+                if ($nStatus <= STATUS_MARKED) {
+                    $zData['class_name'] = ($nStatus == STATUS_MARKED ? 'marked' : 'del');
+                }
+            } else {
```
mkroon1: I tried to make this section a bit more concise and the comments more to the point. Here's what I came up with:

~~~php
            // Determine collective status for all data in current viewlist row.
            $nRowStatus = STATUS_OK;
            if (!LOVD_plus && $_AUTH['level'] >= LEVEL_COLLABORATOR) {
                // Loop through possible status fields.
                foreach (array('statusid', 'var_statusid', 'ind_statusid') as $sField) {
                    if (isset($zData[$sField]) && !empty($zData[$sField])) {
                        // Take minimum value of all status fields.
                        // Note: var_statusid in VariantOnTranscriptUnique view may contain
                        //       multiple IDs in a string. Therefore only the first char is used.
                        $nRowStatus = min($nRowStatus, intval(substr($zData[$sField], 0, 1)));
                    }
                }
            }

            // Determine row's CSS class name based on row status.
            $zData['class_name'] = '';
            if ($nRowStatus == STATUS_MARKED) {
                $zData['class_name'] = 'marked';
            } elseif ($nRowStatus < STATUS_MARKED) {
                $zData['class_name'] = 'del';
            }
~~~

Other than that, I don't see any problems merging as it is.

ifokkema: Thanks! I've used your code (some edits to simplify and keep some of my comments) in 89f4029.

PR #320, opened 2017-12-05
Automated imports using the submission API.

### LINKED ISSUE #149: Create API to submit data into
Creating an API to which data can be sent, allowing automated data import into LOVD3.
- [x] Design authorization schema
- [x] Design format (JSON)
- [x] Implement first step of API
  - [x] Allow users to create and renew authorization token in LOVD3
    - [x] Design token
    - [x] Adapt data model to store token
    - [x] Adapt VE to shown token (hidden, show on request) and provide links
    - [x] When showing token, explain features
    - [x] When creating token while one already exists, warn new token will revoke access of current token
    - [x] Implement creating a token, then forward to showing the token
    - [x] Allow deletion of token, removing all authorization of APIs
    - [x] Let all actions be logged
  - [x] Allow setting data path in config.ini.php file
    - [x] Let API fail when the data path is not configured
  - [x] Implement endpoint to receive data
    - [x] Do we need versioning in the URL?
    - [x] Implement HTTP status codes:
      - [x] HTTP 202 Accepted (accepted for later processing)
      - [x] HTTP 400 Bad Request (malformed syntax)
      - [x] HTTP 401 Unauthorized
      - [x] HTTP 405 Method Not Allowed (when using GET)
      - [x] HTTP 406 Not Acceptable (when response other than plain text or JSON is required)
      - [x] HTTP 413 Payload Too Large (file too large)
      - [x] HTTP 415 Unsupported Media Type (format not supported)
      - [x] HTTP 422 Unprocessable Entity (format OK, syntax OK, semantics wrong)
      - [x] HTTP 500 (other program errors)
      - [x] HTTP 501 Not Implemented (when not using GET or POST)
    - [x] Parse file, generalize it
    - [x] Check authorization
    - [x] Verify minimal data is present
      - [x] Properly check detection_technique field as well, to make sure we don't fail there during actual import.
    - [x] Allow for more transcripts in the file than present in the database; at least one should match, otherwise, throw error
    - [x] Convert the JSON to an LOVD data object
      - [x] We need to calculate all positions for this, and the variant type (#58)
    - [x] Write contents to LOVD import file to data path
    - [x] Write the original JSON file to a file too, for debugging purposes.
- [x] Implement second step of API
  - [x] Implement manual scheduling of files with LOVD+ implementation, but add:
    - [x] Field where errors go
    - [x] Priority flag
  - [x] Add ajax controls:
    - [x] Unschedule
    - [x] Set priority
    - [x] Reimport file (clear in_progress and process_errors fields)
  - [x] Add menu item for the scheduling page.
  - [x] Implement import of scheduled files
    - [x] Archive files? Archiving folder?
    - [x] Report success? LOVD needs this. Data should still be curated.
    - [x] Link to manual run of automated import from scheduling page.
  - [x] Document how to setup this importing with a cron job
- [x] Document format and submission method

Depends on #58.

ifokkema:
Automated imports using the submission API.
- Install scheduled imports data table.
- Added feature to schedule submission files for import.
- Files submitted through the API now have a file extension.
- Added a link from the Setup area and the Setup tab's menu to the page to schedule imports.
- Files submitted through the API are displayed and ranked.
- Icons are shown for various information.
  - A download link to download the file.
  - Information is given about when files have been scheduled and by whom.
  - Whether files are assigned with priority, and the priority name.
  - For files sent through the API, the original filename is stored in the info icon.
  - If files are currently being processed, and the time they started running.
  - If there have been errors, and if so, which errors.
- Added code to automatically import files that have been scheduled.
- Errors encountered are stored in the database, so they are visible on the file scheduler page.
- Some output is now suppressed, because we're trying to create text only output.
  - Import.php no longer sends SCRIPT tags when not outputting HTML.
  - The progress bar now prints the init message to the screen for text/plain output, so the percentages displayed have a more clear meaning.
- Fixing bug; The LOVD receiving a submission might have more custom columns that are mandatory.
  - Try to predict the default values for those, and fill those in.
  - The defaults are now predicted based on the selection_options (needs some work, maybe).
  - Defaults should still be filled in.
  - Added default values for non-shared columns, if it can not be found using the selection lists.
- Added Ajax actions for unscheduling, rescheduling, and setting the priority of a scheduled file.
  - Technology is same as API auth token script.
  - Unscheduling a file is available if the file has been scheduled.
  - Setting the priority is available for files that have not been processed yet. Three levels of priority are available.
  - Rescheduling a file is available for processed files that are still found on disk.
- Keep the JSON file for each API submission.
- Document the watching of the data directory and the automatic import setup, in INSTALL.txt.
- Automated imports can also be run manually.
  - Added an text/html interface.
  - text/plain is only activated when no other format is requested.
  - Added link from the scheduler page (in the info box), when unprocessed scheduled files are available.
  - Added more details to the info box on the scheduler page.
- The autoimporter mails the curators.
  - The autoimporter sends an email to every curator (or the managers in case of intragenic data) that is affected by the imported data, in case non-public data was included.
  - The autoimporter is now only run by the authorization of the LOVD user if you're not already manager.
  - The autoimporter tries to guess (now only for API submissions) which user the data belongs to, so the emails show the correct submitter.
- Added a section in the manual on the submission API and some documentation improvements in the manual.

Closes #149.


src/import.php
```diff
@@ -37,6 +37,346 @@
...
+                                    ORDER BY (sf.processed_date IS NOT NULL) DESC, sf.priority DESC, sf.processed_date, sf.scheduled_date, sf.filename')
+                           ->fetchAllGroupAssoc();
+
+    // [0/1] indicates processed no/yes.
+    // Files in format [filename] = sort_code (unscheduled(0|1), reverse_priority(0-9), processed_date, scheduled_date, file_timestamp).
+    // NOTE: If you change the sort code, be sure to update this comment, the code that builds it, the code the fills in the orphaned database entries, and the scheduling part.
+    $aFiles = array(
+        array(),
+        array(),
+    );
```
mkroon1: I had trouble understanding this format. If I understand correctly, $aFiles is an array where the key is a boolean denoting whether the files listed in the value array are processed or not. And the value itself is an array with filename as key and the value is a sort key in the form of a comma-separated string of the values:
* unscheduled (bool)
* reverse priority (integer between 0 and 9)
* date when processed
* date when scheduled
* date when file was created.

Perhaps an example is more telling:
~~~php
$aFiles = array(
    0 => array(
        'file.txt' => '0,4,2017-01-01 15:10:35,2017-01-01 15:10:36,2017-01-01 15:10:37',
        'file2.txt' => ...
    ),
    1 => ...
);
~~~
But maybe it's better to do away with the sort string altogether (see comment below).

src/import.php
```diff
@@ -37,6 +37,346 @@
...
+    }
+
+
+
+
+
+    // Sort the file list (the DB list is already sorted).
+    asort($aFiles[0]); // Sort data files, keeping indices.
+    asort($aFiles[1]); // Sort data files, keeping indices.
+    $nFilesProcessed = count($aFiles[1]);
```
mkroon1: The code that builds the string to sort files with is IMO hard to understand. I think storing info about the files in $aFiles as separate values and using usort() here with a user-defined comparison function. E.g. this would make hacking the sort value on line 161 more readable (simply change a few values instead of using substr()).

src/import.php
ifokkema: Yes, you understood correctly but you're right this code is complicated. It took me quite some time to build as well :stuck_out_tongue_closed_eyes: 
More on the other comment...

src/import.php
ifokkema: You're right, I should have thought about that. I'll make the changes, it'll take more code but it will make things much simpler to understand if I need to revise this later.

ifokkema: Implemented in e5bf63b. I'll make a few other changes, and will ask you to have another look, if you don't mind.

ifokkema:
Hi @mkroon1 ,

I think I'm done now :wink:

I added some documentation too. If you think this is OK, we can merge it.

PR #321, opened 2017-12-05
View gene homepage by HGNC ID.

### LINKED ISSUE #270: Redirect genes/HGNC_ID to genes/SYMBOL
It would be nice to be able to go directly to a gene via the HGNC ID, this is especially useful when there are more symbols connected to the same gene and it's unknown which one is used in the particular LOVD installation.

The idea is to redirect the page `genes/HGNC_ID` to `genes/SYMBOL`.

mkroon1:
Fixes #270

### DIFF
--- a/src/genes.php
+++ b/src/genes.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-12-15
- * Modified    : 2017-11-20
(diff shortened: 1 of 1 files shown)


ifokkema:
Hi @mkroon1 ,

I had to make some changes here because of some problems with the CURRDB implementation. See my commit's message. If you're OK with this and the test has finished, we can merge it.

PR #331, opened 2018-01-16
Fixed security bug and fixed broken external viewer.

ifokkema:
Fixed security bug and fixed broken external viewer.
- aColsToSkip was now trusted to be set in SESSION, but we did not check if it actually was set in SESSION. The recent removal of code that enforced the removal of certain columns for lower level users, allowed these users to forge an AJAX request that would show the links anyway. Reinstated this code. Luckily, this change had not been released, it was done after the 3.0-20a release.
- aColsToSkip was needed for the external viewer. Not being able to set it anymore, resulted in lots of additional columns for the country node views.

### DIFF
--- a/src/ajax/viewlist.php
+++ b/src/ajax/viewlist.php
@@ -4,10 +4,10 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-02-18
- * Modified    : 2017-11-20
+ * Modified    : 2018-01-16
  * For LOVD    : 3.0-21
  *
- * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
+ * Copyright   : 2004-2018 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
  *               Ivar C. Lugtenburg <I.C.Lugtenburg@LUMC.nl>
  *               Daan Asscheman <D.Asscheman@LUMC.nl>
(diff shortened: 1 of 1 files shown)


src/ajax/viewlist.php
```diff
@@ -121,6 +121,16 @@
     die(AJAX_NO_AUTH);
 }
 
+// Load the columns to skip from the request. The external viewer uses this.
+$aColsToSkip = (!empty($_REQUEST['skip'])? $_REQUEST['skip'] : array());
+
+// Submitters should not be allowed to retrieve more information
+//  about users than the info the access sharing page gives them.
+if ($sObject == 'User' && $_AUTH['level'] < LEVEL_MANAGER) {
```
ifokkema: @mkroon1 , this code is different from what it was in [3.0-20a](https://github.com/LOVDnl/LOVD3/blob/caaaace8c7c95b6354097a169a23fa5bea1eb3c5/src/ajax/viewlist.php#L119). There, it was a check on `LEVEL_SUBMITTER`. However, `LEVEL_CURATOR` is also not allowed to see more columns, right? I don't understand why it was `LEVEL_SUBMITTER`, do you know?

mkroon1: From what I understand of the commit that introduced this check, it was made specifically to block submitters, since those had access to the new viewlist to designate colleagues. I guess collaborators were not considered at the time. I.e.: no specific reason why it didn't say `< LEVEL_MANAGER`.

ifokkema: OK, cool! Collaborators are also kind of submitters, but not stopped by a check on `LEVEL_SUBMITTER`, so it's best to keep the `LEVEL_MANAGER` check then. Thanks!

PR #335, opened 2018-01-26
Lovd light

mkroon1:
Introduces `LOVD_LIGHT` constant in `inc-init.php`. When set to true, this disables a lot of pages/columns/fields througout LOVD, retaining only the parts that are concerned with core variant information.

The features turned off with `LOVD_LIGHT` can be controlled more finely with the values in the global `$_SETT['customization_settings']`.

Setting `LOVD_LIGHT` will certainly break the Selenium tests as they are currently defined.


ifokkema:
Closing, creating new PR instead because the implementation has completely changed.

PR #358, opened 2018-05-29
Added missing viewlist object required by LOVD_external

jplazzer:
- "Transcript,VariantOnTranscript,VariantOnGenome,Screening,Individual"
is used in LOVD_external but wasn't in the available viewlist objects


ifokkema:
Hi @jplazzer ,

Thanks for this reminder! I had this code already implemented in the [shared LOVD instance](https://databases.lovd.nl/shared/), but hadn't put it in LOVD3 yet; simply forgot that the external viewer needed it on other LOVDs as well :sweat_smile: 

I copied that code from there, with the same features but slightly differently implemented, merged in #364, so I'll close this. Thanks again!

PR #380, opened 2019-07-25
Multiple bug fixes.

ifokkema:
Multiple bug fixes.
- Fixed bug; The mapper didn't fill in the VOT's effectid, but left it at its default.
  - Now, it'll take the VOG's effectid and it'll copy it.
- The public gene-specific downloads had some non-public columns included, like status, created_by, created_date, etc.
  - These are now removed. They're not damaging to be shared, but they're not shown on VLs or VEs, so they should remain hidden in the download as well.
- A download will now no longer show inactive Phenotype and VOT columns, if filters are active for Diseases or Genes, respectively.
  - This cleans up the file a lot for single-gene downloads.
  - It reports a warning for Phenotype data, since it's possible that data is selected unrelated to the selected Disease(s).
- When creating a transcript without selecting a gene first, LOVD no longer uses a dropdown but a VL.
  - The dropdown became too large on large databases.
  - The VL is much better equipped to handle large datasets.
  - The output is restricted for curators in case they curate 25 genes or less.
  - The currently selected gene is highlighted.
- Cleaned up the genes?authorize and the users?share_access.
  - Selected user's IDs where stored in both userid and id. Since the latter is visible, part of this selection could be influenced. Now only userid is used, making the filtering of the VL complete invisible to the user.
  - Also, when selecting users as curators/collaborators, their ID wasn't added to their name, like with previously selected users.

### DIFF
--- a/src/ajax/map_variants.php
+++ b/src/ajax/map_variants.php
@@ -4,10 +4,10 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2010-01-15
- * Modified    : 2018-03-09
- * For LOVD    : 3.0-21
+ * Modified    : 2019-07-25
+ * For LOVD    : 3.0-22
  *
- * Copyright   : 2004-2018 Leiden University Medical Center; http://www.LUMC.nl/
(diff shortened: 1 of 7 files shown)


ifokkema:
Not merging yet, awaiting feedback for changes.

PR #387, opened 2019-08-01
Fix `variants/in_gene` view.

### LINKED ISSUE #386: The `variants/in_gene` page is extremely slow on the GV shared.
The `variants/in_gene` page is extremely slow on the GV shared, and it takes almost a minute to render. After some investigation;

- The COUNT(*) is slow because the transcripts, variants_on_transcripts, and variants table are retained in the query.
  - [x] The transcripts table is in there, because it's part of the `GROUP BY`, even though that's not necessary. It's in there to match the `ORDER BY`, but we're already using temp table and filesort anyway, so we might as well drop it from the `GROUP BY`.
  - [x] The VOG and VOT tables are necessary, even though we can optimize a bit (using `INNER JOIN` to VOT and dropping the `WHERE` on `vot.id` will help).
  - [x] After these changes, check the query optimizer.

- The query itself is slow for a few reasons.
  - [x] See above, the `GROUP BY` can be optimized.
  - [x] We are wasting a lot of time with `MIN()` and `MAX()` calls, there because we want to support MySQL instances with `ONLY_FULL_GROUP_BY`. Stop wasting time, and turn off this setting; it's in our way. We're grouping by VOT already, why do all these expensive calculations on VOT and VOG columns...
  - [x] The table joins can be optimized (see above).
  - [x] We sort on `geneid`. Not doing that will make the query return instantly. But I guess we'll have to keep this.

ifokkema:
Fix `variants/in_gene` view.
- Disable SQL mode `ONLY_FULL_GROUP_BY`; it is causing problems.
  - Even if we try to play nice, the totally unnecessary MIN() and MAX() calls slow down queries a lot.
- Removed unneeded `MIN()` and `MAX()` calls from custom viewlists.
- When VOT is joined to Transcripts, use `INNER JOIN`, and no longer group on Transcript columns.
  - This speeds up the query quite a lot.
  - Also, rename `LEFT JOIN`s to `LEFT OUTER JOIN`s to make more clear what they are.

Closes #386.

### DIFF
--- a/src/class/PDO.php
+++ b/src/class/PDO.php
@@ -4,10 +4,10 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2011-08-17
- * Modified    : 2017-01-25
- * For LOVD    : 3.0-19
+ * Modified    : 2019-08-01
+ * For LOVD    : 3.0-22
  *
- * Copyright   : 2004-2017 Leiden University Medical Center; http://www.LUMC.nl/
+ * Copyright   : 2004-2019 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
  *               Ivar C. Lugtenburg <I.C.Lugtenburg@LUMC.nl>
  *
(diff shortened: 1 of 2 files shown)


ifokkema:
`COUNT(*)` cannot be optimized further, as the optimizer would need to be changed too much, and somehow the counts are different in the shared with an optimized query. Also, the instant `in_gene` result only happens when not using a `WHERE` on `vog.statusid`, which is of course needed. So we can't optimize this further.

Issue #550, opened 2021-09-20
Make genome builds configurable


ifokkema:
[Intro]

- [ ] Design a Genome Build table (id, name, column_suffix, created_by, created_date)
- [ ] Create Genome Build table for:
  - [ ] For new installs
  - [ ] For upgraded installs
- [ ] Automatically fill with the standard genome build that has been selected
  - [ ] For new installs
  - [ ] For upgraded installs
- [ ] Remove `refseq_build` from `TABLE_CONFIG`.
  - [ ] Fix dependencies.
- [ ] Remove data duplication from `inc-init.php` and `inc-sql-chromosomes.php`.
  - [ ] Fix dependencies.
- [ ] Allow user to activate a genome build.
  - [ ] Entry must be created in the Genome Build table.
  - [ ] VOG table must be altered (adding start, end, DNA field).
  - [ ] Custom DNA column must be created.
  - [ ] Transcripts table must be altered (adding start, end fields).
- [ ] Allow user to deactivate a genome build.
  - [ ] Entry must be deleted in the Genome Build table.
  - [ ] VOG table must be altered (removing start, end, DNA field).
  - [ ] Custom DNA column must be deactivated.
  - [ ] Transcripts table must be altered (removing start, end fields).
- [ ] Fix dependencies.
- [ ] Add details from multiple genome builds to the API.
- [ ] Allow submission starting any genome build.
  - [ ] Submission form
  - [ ] Submission API
- [ ] Allow import using any genome build

PR #552, opened 2021-09-22
Have genome builds table created, and data viewable

ifokkema:
- Create the genome builds table when installing LOVD.
- Create the genome builds table when upgrading an existing LOVD instance.
- Create the genome builds VL.
- Create the genome builds VE.
- Version bump to 3.5-pre-01.

Related to #550.

### DIFF
--- a/src/class/object_genome_builds.php
+++ b/src/class/object_genome_builds.php
@@ -9,7 +9,7 @@
  *
  * Copyright   : 2004-2021 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
- *               L. Werkman <l.werkman@lumc.nl>
+ *               L. Werkman <L.Werkman@LUMC.nl>
  *
  *
  * This file is part of LOVD.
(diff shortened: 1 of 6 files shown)


src/class/object_genome_builds.php
```diff
@@ -0,0 +1,102 @@
...
+                                    'view' => array('ID', 50),
+                                    'db'   => array('gb.id', 'ASC', true)),
+                        'name' => array(
+                                    'view' => array('name', 50),
+                                    'db'   => array('gb.name', 'ASC', true)),
+                        'created_date' => array(
+                                    'view' => array('created on', 130),
+                                    'db'   => array('gb.created_by', 'DESC', 'DATETIME')),
+                        'created_by_' => array(
+                                    'view' => array('User', 160),
```
ifokkema, line 91: Try to standardize these fields, based on other VLs (ViewLists).

src/class/object_genome_builds.php
```diff
@@ -0,0 +1,102 @@
...
+                                    'db'   => array('u.name', 'ASC', true)),
+                        'column_suffix' => array(
+                                    'view' => array('Column suffix', 100),
+                                    'db'   => array('gb.column_suffix', 'ASC', true))
```
ifokkema, line 95: Also, please standardize the order of columns. `created_date` goes last, `created_by` before that (basically, use the same order of the fields in the database table).

src/class/object_genome_builds.php
```diff
@@ -0,0 +1,102 @@
...
+        // SQL code for viewing a list of entries.
+        $this->aSQLViewList['SELECT']   = 'gb.*, u.name AS created_by_';
+        $this->aSQLViewList['FROM']     = TABLE_GENOME_BUILDS . ' AS gb LEFT OUTER JOIN ' . TABLE_USERS . ' AS u ' .
+                                          'ON (gb.created_by = u.id)';
+        $this->aSQLViewList['ORDER_BY'] = 'id';
+
+        // List of columns and (default?) order for viewing an entry.
+        $this->aColumnsViewEntry =
+            array(
+                'TableHeader_General' => 'General information',
```
ifokkema, line 70: We only use this for the Genes VE, because it contains more tables. Other VEs don't have this. So you can remove this line :)

src/class/object_genome_builds.php
```diff
@@ -0,0 +1,102 @@
...
+                'id' => 'Genome build ID',
+                'name' => 'Genome build name',
+                'column_suffix' => 'Column suffix',
+                'created_by' => 'Created by',
```
ifokkema, line 74: To not show a number but a name, use `created_by_` here (with an underscore).

ifokkema (commented):
Some minor comments on the VL and VE definitions.

PR #556, opened 2021-09-28
Gb/enable

### LINKED ISSUE #553: Add/enable new GB
- [x] Allow user to add new GB (`/genome_builds?add`)
  - [x] Give form to add new GB
     - [x]  Show list of all possible GBs (only show GBs of the currently active organism)
     - [x] Make sure only one new GB is added per form
  - [x] Add row in GB table
  - [x] Register custom DNA column
  - [x] Add columns in VOG table (start, end, DNA)
  - [x] Add columns in Transcripts table (start, end)

Related to #550

loeswerkman:
This code makes sure that the necessary columns are added to the VOG and Transcripts tables after enabling a new GB. 

Closes #553 
Related to #550

### DIFF
--- a/src/genome_builds.php
+++ b/src/genome_builds.php
@@ -4,8 +4,8 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2021-09-21
- * Modified    : 2021-09-22
- * For LOVD    : 3.5-pre-01
+ * Modified    : 2021-09-27
+ * For LOVD    : 3.5-pre-02
  *
  * Copyright   : 2004-2021 Leiden University Medical Center; http://www.LUMC.nl/
  * Programmers : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
@@ -81,8 +81,8 @@ if (PATH_COUNT == 1 && ACTION == 'add') {
     if (POST) {
         lovd_errorClean();
 
-        if (empty($_POST)) {
-            // Raise error if ID is empty or not addable
+        if (empty($_POST['id'])) {
+            // Raise error if ID is empty or not addable.
             lovd_errorAdd('id', 'Please select a genome build.');
         }
 
@@ -102,8 +102,7 @@ if (PATH_COUNT == 1 && ACTION == 'add') {
             // Add new genome build as new row into GenomeBuilds table.
             $_DATA->insertEntry($_POST, $aFields);
 
-
-            // Add custom DNA column into VOG table.
+            // Register the new DNA column for the VOG table (TABLE_COLS and TABLE_ACTIVE_COLS).
             $aQueries = array_slice(lovd_getActivateCustomColumnQuery(array('VariantOnGenome/DNA')), 0, 2);
             $aQueries = array_map(function ($s) {
                 return str_replace(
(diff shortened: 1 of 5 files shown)


src/genome_builds.php
```diff
@@ -85,9 +97,11 @@
             // Raise error if ID is empty or not addable.
             lovd_errorAdd('id', 'Please select a genome build.');
         }
-
-        // TODO: Ask Ivo how to use var $aAddableGBs up OR how to use $_POST below,
-        // TODO: to be able to throw an error when a user tries to add an unaddable GB
+        elseif (in_array($_POST['id'], $aAddableGenomeBuilds)) {
```
ifokkema: This check isn't correct indeed, because the ids (e.g., "hg38") are the *keys* of `$aAddableGenomeBuilds`. However, `in_array()` doesn't check the keys, but the values. I updated this and used `isset()` here to check for the existence of the key.

src/genome_builds.php
```diff
@@ -124,46 +141,69 @@
...
+                    'position_g_mrna_end_' . $_POST['column_suffix']))->fetchAllColumn();
+
+            $aTablesAndTheirColumns = array(
+                TABLE_VARIANTS => array(
+                    '`VariantOnGenome/DNA/' . $_POST['column_suffix'] . '`' => array(
+                        'type' => 'VARCHAR(255)',
+                        'active' => in_array(
+                            'VariantOnGenome/DNA/' . $_POST['column_suffix'],
+                            $aActiveColumnsVariants)
+                    ),
```
ifokkema, line 164: This still has a lot of repetition, and it's still quite long. I will update this to show a much shorter alternative.

src/genome_builds.php
```diff
@@ -124,46 +141,69 @@
...
+                        'type' => 'INT(10) UNSIGNED AFTER position_g_end',
+                        'active' => in_array('position_g_start_' . $_POST['column_suffix'], $aActiveColumnsVariants),
+                    ),
+                    'position_g_end_' . $_POST['column_suffix'] => array(
+                        'type' => 'INT(10) UNSIGNED AFTER position_g_start_' . $_POST['column_suffix'],
+                        'active' => in_array('position_g_end_' . $_POST['column_suffix'], $aActiveColumnsVariants)
+                    ),
+                ),
+                TABLE_TRANSCRIPTS => array(
+                    // TODO: Ask Ivo whether or not the mrna positions should be NOT NULL
```
ifokkema, line 175: You're right, they can't be "NOT NULL", as we currently don't have values for them. So the default (which allows `NULL`) is good.
By the way - you can add todos to the issue or the PR, so they don't have to be stored in the code.

ifokkema (commented):
I have highlighted some parts where I either already made changes or where I will still make changes. Please check my comments and my commits; if it's unclear why I did something, please let me know!

ifokkema:
I updated the PR to point to `release/3.5` instead of `master` :wink:

src/genome_builds.php
```diff
@@ -132,68 +132,37 @@
...
-                    'position_g_mrna_start_' . $_POST['column_suffix'],
-                    'position_g_mrna_end_' . $_POST['column_suffix']))->fetchAllColumn();
-
+            // Prepare an array to more easily add the required columns into the
+            //  VOG and transcripts tables.
             $aTablesAndTheirColumns = array(
                 TABLE_VARIANTS => array(
-                    '`VariantOnGenome/DNA/' . $_POST['column_suffix'] . '`' => array(
-                        'type' => 'VARCHAR(255)',
-                        'active' => in_array(
```
ifokkema: This is too much repetition, so I made this check dynamic.

src/genome_builds.php
```diff
@@ -132,68 +132,37 @@
...
-                        'type' => 'INT(10) UNSIGNED AFTER position_g_mrna_end',
-                        'active' => in_array('position_g_mrna_end_' . $_POST['column_suffix'], $aActiveColumnsTranscripts)
-                    ),
+                    'position_g_mrna_start_' . $_POST['column_suffix'] => 'INT(10) UNSIGNED AFTER position_g_mrna_end',
+                    'position_g_mrna_end_' . $_POST['column_suffix'] => 'INT(10) UNSIGNED AFTER position_g_mrna_start_' . $_POST['column_suffix'],
                 ),
             );
 
-            // Add columns to VOG and Transcripts tables.
-            foreach (array_keys($aTablesAndTheirColumns) as $sTable) {
```
ifokkema: Please check the [PHP manual on foreach](https://www.php.net/manual/en/control-structures.foreach.php); if you need both the keys and the values in an array, use:
```php
foreach ($aArray as $sKey => $aVal) {
    ...
}
```
which is much more efficient and keeps the code more readable as well.

src/genome_builds.php
```diff
@@ -132,68 +132,37 @@
...
+                      AND COLUMN_NAME IN (' . implode(',', array_fill(0, count($aTable), '?')) . ')',
+                    array_merge(array($sTable), array_keys($aTable)))->fetchAllColumn();
+
+                $bToAdd = false;
                 $sSQL = 'ALTER TABLE ' . $sTable;
 
-                foreach (array_keys($aTablesAndTheirColumns[$sTable]) as $sColumn) {
-                    $aColumn = $aTablesAndTheirColumns[$sTable][$sColumn];
-                    if (!$aColumn['active']) {
-                        $bToAdd = True;
```
ifokkema: `True` and `False` feel a bit "Python-ish", and can simply be written as `true` and `false`.

ifokkema (commented):
I hightlighted some of the changes that I made in this commit. Please check if they all make sense to you; if not, please ask!

PR #558, opened 2021-09-30
Gb/disable

### LINKED ISSUE #557: Allow users to deactivate a genome build
- [x] Add button to the GB ViewEntry (VE) to deactivate the genome build (like done on other VEs)

- [x] Show form to deactive GB (`/genome_builds/hg19?remove`)
  - [x] Just give a warning and require a password, no other fields needed.

- [x] Give an error and block removal when:
   - [x] The chosen ID is not among the active IDs
   - [x] Password is incorrect
   - [x] There are variants on the chosen genome build that are not yet mapped on one of the other active genome builds

- [x] Delete entry if all variants are safely mapped on another genome build
   - [x] Delete row from Genome build table
   - [x] Alter VOG table (deleting the start, end, and DNA columns)
   - [x] Deactivate the custom DNA column
   - [x] Alter Transcripts table (deleting the start and end columns)

Related to #550

loeswerkman:
Allows users to deactivate a given genome build.
- Adds option to deactivate in VE of genome builds, ONLY if there is more than one genome build active.
- Checks to make sure all variants are safely stored on remaining genome builds, and cancels the removal when this is not the case.

Closes #557 
Related to #550

### DIFF
--- a/src/genome_builds.php
+++ b/src/genome_builds.php
@@ -371,8 +371,8 @@ if (PATH_COUNT == 2 && ACTION == 'remove') {
             }
 
             // Deactivate custom DNA column.
-            $_DB->query('DELETE FROM ' . TABLE_ACTIVE_COLS .
-                ' WHERE colid = "VariantOnGenome/DNA' . $sSuffixWithSlash . '"');
+            $_DB->query('DELETE FROM ' . TABLE_ACTIVE_COLS . ' WHERE colid = ?',
+                array('VariantOnGenome/DNA' . $sSuffixWithSlash));
 
             // Write to log...
             lovd_writeLog('Event', LOG_EVENT, 'Removed Genome Build ' . $sID);
--- a/src/genome_builds.php
+++ b/src/genome_builds.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2021-09-21
- * Modified    : 2021-09-28
+ * Modified    : 2021-09-30
  * For LOVD    : 3.5-pre-02
  *
  * Copyright   : 2004-2021 Leiden University Medical Center; http://www.LUMC.nl/
@@ -68,7 +68,7 @@ if (PATH_COUNT == 1 && !ACTION) {
 
 if (PATH_COUNT == 1 && ACTION == 'add') {
     // URL: /genome_builds?add
-    // Enable a new genome builds.
+    // Enable a new genome build.
 
     define('PAGE_TITLE', lovd_getCurrentPageTitle());
     define('LOG_EVENT', 'GenomeBuildAdd');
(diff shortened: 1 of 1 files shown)


src/genome_builds.php
```diff
@@ -244,7 +244,193 @@
...
+
+        if (empty($_POST['password'])) {
+            lovd_errorAdd('password', 'Please fill in the \'Enter your password for authorization\' field.');
+
+        } elseif (!lovd_verifyPassword($_POST['password'], $_AUTH['password'])) {
+            lovd_errorAdd('password', 'Please enter your correct password for authorization.');
+
+        } else {
+            $bValidPassword = true;
+        }
```
ifokkema: Better stick to the LOVD defaults here;
- `lovd_errorClean()` cleans up any previous errors if there are any (almost never the case, but still).
- `lovd_errorAdd()` adds an error.
- `lovd_error()` then returns whether or not there are errors.
This way, you won't have to make your own variables.

src/genome_builds.php
```diff
@@ -244,7 +244,193 @@
...
+            exit;
+        }
+    }
+
+
+
+    $_T->printHeader();
+    $_T->printTitle();
+
+    $sName = $sID . ' / ' . $_SETT['human_builds'][$sID]['ncbi_name'];
```
ifokkema, line 400: Better take this out of the database instead of redefining how we name a genome build. If we'd ever change how we name them, we'd need to change this code as well, or it'll mention a name that may not be used in the database.
(I'll fix this in a new commit)

src/genome_builds.php
```diff
@@ -244,7 +244,193 @@
...
+
+    lovd_showInfoTable('This will deactivate genome build ' . $sName . ' from your database.', 'warning');
+
+    lovd_errorPrint();
+
+    // Tooltip JS code.
+    lovd_includeJS('inc-js-tooltip.php');
```
ifokkema: This is not wrong, but simply unneeded since this form doesn't have any tooltips. So I chose to remove it to prevent a file from getting included that wasn't used at all.

src/genome_builds.php
```diff
@@ -384,8 +384,9 @@
             }
 
             // Deactivate custom DNA column.
+            // TODO: Colsuffix / dependant
             $_DB->query('DELETE FROM ' . TABLE_ACTIVE_COLS .
-                ' WHERE colid = "VariantOnGenome/DNA/' . $sID . '"');
+                ' WHERE colid = "VariantOnGenome/DNA' . $sSlash . $aActiveBuilds[$sID] . '"');
```
ifokkema: Better use PDO placeholders here (`?`). I'll fix this.

ifokkema (changes requested):
Some changes required; I'll commit them separately.

PR #559, opened 2021-10-01
Percentage complete

loeswerkman:
Added a percentage bar to the VE give an idea of the completion of the genome build.

Related to #550

### DIFF
--- a/src/class/object_genome_builds.php
+++ b/src/class/object_genome_builds.php
@@ -135,10 +135,10 @@ class LOVD_GenomeBuild extends LOVD_Object
             $sCompletionBar = '' .
                 '      <TABLE border="0" cellpadding="0" cellspacing="0" width="200"' .
                 '        <TR>' .
-                '          <TD width="200" style="border : 1px solid black; padding : 0px;">' .
+                '          <TD width="200" style="border : 1px solid black; padding : 0px; height=10px">' .
                 '            <IMG src="gfx/trans.png" alt="" width="' . $iPercentComplete . '%" height="11"' .
                 '             style="background : #224488;"></TD>' .
-                '          <TD id="lovd_progress_value" style="font-size:10px">' . $iPercentComplete . '%</TD>' .
+                '          <TD id="lovd_progress_value" style="font-size:11px">' . $iPercentComplete . '%</TD>' .
                 '        </TR>' .
                 '      </TABLE>';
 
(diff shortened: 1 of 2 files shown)


src/class/object_genome_builds.php
```diff
@@ -115,6 +116,33 @@ function prepareData ($zData = '', $sView = 'list')
...
             $zData['created_date_'] = substr($zData['created_date'], 0, 10);
+
+        } elseif ($sView == 'entry') {
+            global $_DB;
+
+            $sSlash = ($zData['column_suffix']) ? '/' : '';
+
+            $iPercentComplete = $_DB->query(
+                'SELECT (mapped_variants / all_variants) * 100 FROM
+                    (SELECT COUNT(id) AS mapped_variants FROM lovd_variants
```
ifokkema, line 127: Note that this query contains a hardcoded table name. Luckily, our table prefixes differ, and I got a query error :sweat_smile: 
I'll fix it when I move the query and merge it with the VE's query.

src/class/object_genome_variants.php
```diff
@@ -367,7 +367,7 @@ function getForm ()
                         array('', '', 'print', '<B>Genomic variant information</B>'),
                         'hr',
                         array('Allele', '', 'select', 'allele', 1, $aSelectAllele, false, false, false),
-                        array('', '', 'note', 'If you wish to report an homozygous variant, please select "Both (homozygous)" here.'),
+                        array('', '', 'note', 'If you wish to report a homozygous variant, please select "Both (homozygous)" here.'),
```
ifokkema, line 370: Thanks :sweat_smile: :smile:

ifokkema (changes requested):
I will make small changes before merging this PR.

src/class/object_genome_builds.php
```diff
@@ -115,6 +116,33 @@ function prepareData ($zData = '', $sView = 'list')
...
+                     WHERE `VariantOnGenome/DNA' . $sSlash . $zData['column_suffix'] . '` IS NOT NULL) mv,
+                    (SELECT COUNT(id) AS all_variants FROM lovd_variants) v')->fetchAllColumn()[0];
```
ifokkema, line 129: A `fetchAllColumn()` followed by a `[0]` is much less efficient than just a `fetchColumn()`.

Issue #573, opened 2021-11-12
Create a batch HGVS checker on LOVD.nl


ifokkema:
Depends on #566.

The internal LOVD function `lovd_getVariantInfo()` is currently being rebuilt by @loeswerkman which adds detailed HGVS checks and informative error messages in the case that variants are not HGVS-conformant. It would be a waste to not use this code also more broadly. Especially since tools like VariantValidator and Mutalyzer seem to sometimes give very random-looking and unclear error messages, it would be beneficial to create a batch HGVS syntax checker for our LOVD website using our new function that provides useful error messages to guide users in the right direction to fix their non-HGVS variant descriptions.

- [ ] Create a good set of correct and incorrect variant descriptions (e.g., our `lovd_getVariantInfo()` unit test set).
- [ ] Run this set through the Variant Validator batch checker.
- [ ] Run this set through the Mutalyzer batch checker.
- [ ] Compare our error messages with theirs. Document this well, so we can use it for a detailed report.
- [ ] See where we can improve if needed.
- [ ] Write a batch checker for LOVD.nl that does better. May require a separate issue or more details in this issue.
- [ ] Will be worth disseminating on social media and the HUGO newsletter.

PR #587, opened 2022-03-03
Feat/check hgvs tool

### LINKED ISSUE #575: Batch syntax / variant checker on LOVD.nl
Depends on #573.

Create an interface that:
- [x] Works on the website. Since the LOVD.nl code is not (yet) versioned, either develop in LOVD's `scripts` folder or create a private LOVD.nl repository. Probably the former.
- [x] Allows both syntax and variant validation, to be chosen by the user. Syntax check is fast, variant validation will be slower.
  - [x] Syntax validation is always done first. When we have validated (and possibly improved) the syntax, the variant may also be sent to VV, which can in turn actually validate the variant itself. VV may also correct the description. Note that for variant validation, a reference sequence is required.
  - [ ] If users submit LRGs or NGs and they want full validation of their variants, then we ought to use the VV endpoint and not the LOVD endpoint.
- [x] Allows both for one-variant checks as well as a list of variants.
  - [x] The list may be populated by copy/paste or the upload of a document.
- [x] Results should be written to the screen, but should also become downloadable.
- [x] Results should include;
  - [x] The input
  - [x] The corrected output
  - [x] Whether there was a problem or not
  - [x] Detailed error messages.
- [x] It would be good to study the VV output format, so we can align our formats.
- [x] Perhaps, due to time constraints, it would be good to first have the syntax checker working and online, and work on the rest after.

loeswerkman:
Add an visual HGVS syntax checker tool using `lovd_getVariantInfo()` and `lovd_fixHGVS()`.
- This branch builds the HGVS syntax checker: an online form that takes a variant and finds out whether it seems to be described following the HGVS nomenclature.
- The form accepts either a single variant input, or a batch variant input, where a list of variant can be pasted into the input form.
- In both cases, it checks the variants' syntax and suggests a correction when possible. It also allows to validate variants using the [VariantValidator](https://www.VariantValidator.org) service. When generating a suggested fix ourselves, we express our level of confidence that this correction is indeed as intended, in the text of the suggestion.
- The interface shows all results below the input, allows a quick method to accept corrections of either `lovd_fixHGVS()` or the VariantValidator service, and allows downloading the results to a text file so that the user can store the result of the analysis.
- The old `ajax/check_hgvs.php` script got merged into the new `ajax/check_HGVS.php` since they have similar functionality. Removed the Mutalyzer API call from the old code and replaced it with the call to LOVD's `lovd_getVariantInfo()`. We provide more accurate syntax checking, including semantic validation.
- Added tests that check the form's functionality completely.

Also:
- Logic concerning reference sequences in variant descriptions has been isolated into separate functions.
- Both `lovd_getVariantInfo() and `lovd_fixHGVS()` have been improved.
  - A specific error is returned when a RefSeq ID is missing its version number.
  - A bug has been fixed which was found in the handling of warnings concerning substitutions with multiple positions.
  - Recognize and fix swapped reference sequences (e.g., `NM_123456.1(NC_123456.1):c.100del`).
  - Use the info from `lovd_getVariantInfo()` to fix `WREFERENCEFORMAT`.
  - Improve `lovd_getVariantInfo()` by adding the 'range' boolean.
  - Fix notice in `lovd_fixHGVS()` triggered by `g.100DEL`. Also added a fix for this type of mistake.
  - Correct uncareful suffix fixing in `lovd_fixHGVS()`.
  - Replace another `str_replace()` with a more intelligent solution.
  - Fix infinite recursion in `lovd_fixHGVS()`.
  - Add tests for new features in `lovd_fixHGVS()` and synced tests with those for `lovd_getVariantInfo()`.
  - Fix typo common to some warning messages from `lovd_getVariantInfo()`.
  - Fix variants with an incorrect order in their inserted sequence.
  - Improve `lovd_getVariantInfo()` to recognize RefSeqs without underscores.
  - Improve `lovd_fixHGVS()` to handle RefSeq errors better.
- Stop recognizing del/dup with a suffix as valid on the data submission form; our functions would not allow it either.
- The VV library has been improved.
  - Simplified Variant Validator library a bit. `lovd_getVariantInfo()` has been improved, so less work is needed in this library.
  - Reformat the VV library, no code changes. Removed redundant nesting that also inhibits quick interpretation of the code.
  - When VV runs into issue [421](https://github.com/openvar/variantValidator/issues/421), call VV again, but just once.
- Checked the project for CRLF and changed it to LF.
- Fix URL parsing issues with URL encoded `?` and arguments with `../`. 
- Synced textual changes and additional comments with the [`api.lovd.nl`](https://github.com/LOVDnl/api.lovd.nl) project.
  - The GPL license text now uses https, and one URL was updated.
  - Some formatting changes to errors thrown in `/src/class/api.php`.
  - The GA4GH API now has some additional info on the variables.
  - Fixed `lovd_arrayInsertAfter()`, which contained some issues.
  - Move the submission API's `jsonDecode()` method to the `LOVD_API` object.

Closes #575 and is related to #573.

### DIFF
--- a/src/inc-lib-init.php
+++ b/src/inc-lib-init.php
@@ -4,7 +4,7 @@
  * LEIDEN OPEN VARIATION DATABASE (LOVD)
  *
  * Created     : 2009-10-19
- * Modified    : 2022-08-05
+ * Modified    : 2022-08-26
  * For LOVD    : 3.0-29
  *
  * Copyright   : 2004-2022 Leiden University Medical Center; http://www.LUMC.nl/
@@ -38,7 +38,8 @@ $_LIBRARIES = array(
         'refseq' => array(
             'basic' => '/^[A-Z_.t0-9()]+$/',
             'strict'  =>
-                '/^([NX][CGMRTW]_[0-9]{6,9}\.[0-9]+' .
+                '/^([NX][CGMRTW]_[0-9]{6}\.[0-9]+' .
+                '|[NX][MR]_[0-9]{9}\.[0-9]+' .
                 '|N[CGTW]_[0-9]{6}\.[0-9]+\([NX][MR]_[0-9]{6,9}\.[0-9]+\)' .
                 '|ENS[TG][0-9]{11}\.[0-9]+' .
                 '|LRG_[0-9]+(t[0-9]+)?' .
@@ -57,21 +58,24 @@ $_LIBRARIES = array(
 
 
 
-function lovd_arrayInsertAfter ($key, array &$array, $new_key, $new_value)
+function lovd_arrayInsertAfter ($sKey, &$a, $sKeyToInsert, $ValueToInsert)
 {
-    // Insert $new_key, $new_value pair after entry $key in array $array.
-    // Courtesy of http://eosrei.net/
-    if (array_key_exists($key, $array)) {
-        $new = array();
-        foreach ($array as $k => $value) {
-            $new[$k] = $value;
-            if ($k === $key) {
-                $new[$new_key] = $new_value;
+    // Insert $sKeyToInsert having $ValueToInsert,
+    //  after entry $sKey in array $aOri.
+    // Based on code by Brad Erickson (http://eosrei.net/comment/287).
+    // MIT licensed code, compatible with GPL.
+    if (array_key_exists($sKey, $a)) {
+        $aNew = array();
+        foreach ($a as $k => $value) {
+            $aNew[$k] = $value;
+            if ($k === $sKey) {
+                $aNew[$sKeyToInsert] = $ValueToInsert;
             }
         }
-        return $new;
+        $a = $aNew;
+        return true;
     }
-    return FALSE;
+    return false;
 }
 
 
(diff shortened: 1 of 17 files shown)


src/inc-lib-init.php
```diff
@@ -2837,6 +2837,17 @@ function lovd_queryError ($sErrorCode, $sSQL, $sSQLError, $bHalt = true)
 
 
 
+function lovd_removeRefSeq ($sVariant)
+{
+    // This function removes the reference sequence from
+    //  a variant description.
+    return substr(strstr($sVariant, ':'), 1);
+}
```
ifokkema, line 2845: Do you think this function should do something "special" when the input has no `:`?

The "downside" of creating such a function is that you may need input validation, especially when notices or warnings could be generated by providing invalid input. In this case, sending an empty string or a boolean doesn't seem to cause problems. But sending in an array does. Better check if `$sVariant` is a string before doing anything? Also, for instance, do you think it's acceptable that running `lovd_removeRefSeq()` on `c.100del` returns an empty string?

You wanted a function, these are the consequences :smiling_imp:

ifokkema (changes requested):

ifokkema (approved):

PR #588, opened 2022-03-03
Improve/variant creation form

### LINKED ISSUE #562: Implement VariantValidator for all genome builds in variant data entry form
- [x] Mark all DNA fields with their genome build (instead of only the first one)
- [x] Allow user to map from whichever DNA field that is in the form
  - [x] If the organism of choice is human:
    - [x] Make sure the HGVS check works on each field (apply ONLY regular expression on any gb that is not hg19 or hg38)
    - [x] Automatically fill all other DNA fields
  - [x] Fill in the `source`
- [x] In `inc-js-variants.php`, get rid of the usage of `$_GET['chromosome']` if you can (can we use the new data elements feature for this?)

loeswerkman:
Update the variantCreationForm to enforce mapping and clean HGVS.

This branch builds the checks and the dialogue which make sure the user inserts clean variants into the database. The dialogue serves as a communication tool throughout the whole process: it will inform the user of steps failing or passing, or of any irregularities along the way. This way, we hope to be able to enforce clean HGVS input as much as possible, without it having to cost us user friendliness. All checks are firstly based on `getVariantInfo` and `fixHGVS` to check the syntax, followed by a call to VariantValidator for a validation of the positions and to retrieve the mapping.

Closes #562 and is related to #550.


loeswerkman:
Issue #562 mentions wanting to remove the usage of $_GET['chromosome'] in inc-js-variants.php. By doing so, we realised it's part of a JS function which is no longer in use: `lovd_convertPosition`. This function was used to convert positions using Mutalyzer, but is not used anymore, since we are now using VariantValidator instead. However, we might want to hold onto this function for the time being, since Mutalyzer could convert positions of mitochondrial DNA whereas VariantValidator, as of yet, cannot.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -43,15 +43,12 @@
...
-    print('
-    // Resetting the md5 key.
-    $(\'input[name="codedVariants"]\').val("");
-    
+// Reset all values.
+print('
+// Resetting the md5 key.
+$(\'input[name="codedVariants"]\').val("");
+
+if ($(\'#variantForm input[name*="VariantOn"]\').hasClass("acc")) {
```
loeswerkman, line 49: This is THE OPPOSITE of what was said in the commit message: so the code here shows that we **are only resetting all values if there ARE values of the class `acc`epted**. The commit message implies that we would only reset "IF there were no variants successfully mapped". The issue here lies within the commit message (and me writing it wrongly). So, to elaborate: we **only want to reset if any values were set to `acc`epted**, since these values were mapped by VariantValidator and these are the values which need reconsideration after a value has been changed. We do NOT need to reconsider variants which were never changed by us in the first place; these are being checked thoroughly after the submit, so we don't need to worry about them.

My apologies for the confusing commit message here :).

src/variants.php
```diff
@@ -833,18 +833,8 @@ function(sData) {
...
-                //  be a genome build (from VOG/DNA/hg38) or "DNA" (from VOG/DNA).
-                $sColumnSuffix = ($_POST['source'] == 'DNA'? '' : $_POST['source']);
-
-                // Get the ID by its column suffix and give this as the source.
-                $sID = $_DB->query(
-                    'SELECT id FROM ' . TABLE_GENOME_BUILDS . '
-                     WHERE column_suffix = ?', array($sColumnSuffix))->fetchColumn();
-                $_POST['source'] = $sID;
+            } elseif (!$_DB->query('SELECT COUNT(*) FROM ' . TABLE_GENOME_BUILDS . ' WHERE ID = ?', array($_POST['source']))->fetchColumn()) {
+                unset($aFieldsGenome[array_search('source', $aFieldsGenome)]);
```
loeswerkman, line 837: Find out whether this was supposed to be here

src/inc-lib-form.php
```diff
@@ -891,7 +891,7 @@ function lovd_viewForm ($a,
      * 'end_fieldset',
```
loeswerkman, line 891: Why was this done again?

src/class/object_genome_variants.php
```diff
@@ -351,9 +355,15 @@ function getForm ()
             }
         }
 
-        // Add genome build name to VOG/DNA field.
-        $this->aColumns['VariantOnGenome/DNA']['description_form'] = '<B>Relative to ' . $_CONF['refseq_build'] . ' / ' . $_SETT['human_builds'][$_CONF['refseq_build']]['ncbi_name'] . '.</B>' .
-            (!$this->aColumns['VariantOnGenome/DNA']['description_form']? '' : '<BR>' . $this->aColumns['VariantOnGenome/DNA']['description_form']);
+        // Add genome build name and element data to VOG/DNA fields, and add in the link to the HGVS website.
+        $aActiveBuilds = $_DB->query('SELECT id, name, column_suffix FROM ' . TABLE_GENOME_BUILDS)->fetchAllGroupAssoc();
```
loeswerkman, line 359: Add the 'Relative to <genome_build>' message in the Variant Creation Form.

src/class/object_genome_variants.php
```diff
@@ -268,6 +254,24 @@ function checkFields ($aData, $zData = false, $aOptions = array())
             }
         }
 
+        foreach ($aData as $sField => $sVariant) {
```
loeswerkman, line 257: Check the HGVS of all DNA variants.

src/styles.css
```diff
@@ -108,7 +108,7 @@ table.error td {padding : 0px 4px;}
```
loeswerkman, line 108: Add CSS class for accepted variants.

src/scripts/fix_variant_descriptions.php
```diff
@@ -549,7 +549,7 @@ function ($sVOT)
                                     } elseif ($aVOT['RNA'] == str_replace('?', '', $aVVVot['data']['RNA'])
```
loeswerkman, line 549: Call isHGVS() instead of getVariantInfo()

src/inc-lib-variants.php
```diff
@@ -108,7 +108,7 @@ function lovd_fixHGVS ($sVariant, $sType = '')
     $sVariant = str_replace(array('‐', '−', '–', '—'), '-', $sVariant);
```
loeswerkman, line 108: Call isHGVS() instead of getVariantInfo()

src/inc-lib-init.php
```diff
@@ -2369,6 +2369,18 @@ function lovd_isColleagueOfOwner ($sType, $Data, $bMustHaveEditPermission = true
```
loeswerkman, line 2369: Build isHGVS() wrapper function.

src/inc-init.php
```diff
@@ -360,6 +360,7 @@
                             'hg18' =>
```
loeswerkman, line 360: Add supported_by_VV in $_SETT array holding all human_builds

src/class/variant_validator.php
```diff
@@ -882,7 +882,7 @@ public function verifyVariant ($sVariant, $aOptions = array())
             //  NM, we end up with only one NM here.
```
loeswerkman, line 882: Fix typo.

src/class/object_genome_variants.php
```diff
@@ -268,6 +254,24 @@ function checkFields ($aData, $zData = false, $aOptions = array())
...
+                if (!lovd_isHGVS($sVariant)
+                    && !isset($aVariantInfo['errors']['ENOTSUPPORTED']) // Supported by LOVD.
+                    && !(isset($aVariant['warnings']['WNOTSUPPORTED'])  // Supported by VariantValidator.
+                        || isset($aVariant['messages']['IUNCERTAINPOSITIONS'])
+                        || isset($aVariant['messages']['IPOSTIONRANGE']))) {
+                    lovd_errorAdd($sField, 'The variant ' . $sVariant . ' did not pass our checks. Please take another look and try again.');
+                }
+            }
+        }
+
```
loeswerkman, line 274: Check the HGVS of all DNA input variants.

src/class/object_genome_variants.php
```diff
@@ -351,9 +355,15 @@ function getForm ()
...
-            (!$this->aColumns['VariantOnGenome/DNA']['description_form']? '' : '<BR>' . $this->aColumns['VariantOnGenome/DNA']['description_form']);
+        // Add genome build name and element data to VOG/DNA fields, and add in the link to the HGVS website.
+        $aActiveBuilds = $_DB->query('SELECT id, name, column_suffix FROM ' . TABLE_GENOME_BUILDS)->fetchAllGroupAssoc();
+        foreach ($aActiveBuilds as $sBuild => $aBuild) {
+            $sColumn = 'VariantOnGenome/DNA' . (!$aBuild['column_suffix']? '' : '/' . $aBuild['column_suffix']);
+            $this->aColumns[$sColumn]['description_form'] = '<B>Relative to ' . $aBuild['name'] . '.</B>' .
+                (!$this->aColumns[$sColumn]['description_form']? '' : '<BR>' . $this->aColumns[$sColumn]['description_form']);
+            $this->aColumns[$sColumn]['element_data'] = array('genome_build' => $sBuild);
+            $this->aColumns[$sColumn]['form_type'][0] = str_replace('(HGVS format)', '(<A href="http://varnomen.hgvs.org/recommendations/DNA" target="_blank">HGVS format</A>)', $this->aColumns[$sColumn]['form_type'][0]);
+        }
```
loeswerkman, line 366: Add the 'Relative to <genome_build>' message in the Variant Creation Form, plus a link to varnomen.HGVS.org .

src/class/object_genome_builds.php
```diff
@@ -123,7 +123,7 @@ function prepareData ($zData = '', $sView = 'list')
             $sDNAColumn = 'VariantOnGenome/DNA' . (!$zData['column_suffix']? '' : '/') . $zData['column_suffix'];
```
loeswerkman, line 123: Fix typo.

src/ajax/mobidetails.php
```diff
@@ -111,10 +111,14 @@
     $_SESSION['csrf_tokens']['mobidetails_confirm'] = md5(uniqid());
```
loeswerkman, line 111: Add documentation stating that flush() doesn't work; Suppress messages when ob_end_flush() fails by adding a '@'.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+                'predict_protein'    => true,          // Should we get protein predictions?
+                'lift_over'          => false,         // Should we get other genomic mappings of this variant?
+                'select_transcripts' => array_diff($aTranscripts, array($sReferenceSequence)), // Should we limit our output to only a certain set of transcripts?
+            ))
+        );
+
+        if (!isset($aMappedViaGB['data']['transcript_mappings'])) {
+            // If for any reason no genomic mappings were given, we cannot perform this
+            //  extra step, and will thus miss some information. We will inform the user.
+            update_dialogue('<br>Your variant could not be mapped to all transcripts due to unknown issues.');
```
loeswerkman, line 504: An issue might lie here:
If a variant IS mapped to some extend, but there are fields which are left empty, the user has no way of filling these fields (since this script resets all mapped values when it is called again).

src/class/object_genome_variants.php
```diff
@@ -222,20 +222,6 @@ function __construct ()
...
-        if (isset($aForm[$sPrefix . 'VariantOnGenome/DNA'])) {
-            $aForm[$sPrefix . 'VariantOnGenome/DNA'][0] = str_replace('(HGVS format)', '(<A href="http://varnomen.hgvs.org/recommendations/DNA" target="_blank">HGVS format</A>)', $aForm[$sPrefix . 'VariantOnGenome/DNA'][0]);
-        }
-        return $aForm;
-    }
-
-
-
-
-
```
loeswerkman, line 238: @ifokkema, this was removed in October (see [commit](https://github.com/LOVDnl/LOVD3/commit/1d3dad73e57a6ba4adeea040b616ef0c8083eaee)). Do you remember why this was deleted?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+    //  on a VOT creation form, and there are multiple transcripts open,
+    //  we want each transcript to get a mapping. To get this, we then
+    //  need to call VariantValidator a second time using one of the
+    //  genomic variants as were returned using our first call.
+    if ($sType == 'VOT' && count($aTranscripts) > 1) {
+
+        $aMappedViaGB = (
+        !isset($aMappedVariant['data']['genomic_mappings']['hg38']) ? // Yes=We have a genomic reference from our first call; No=We don't have a genomic reference.
+            array() :
+            $_VV->verifyGenomic($aMappedVariant['data']['genomic_mappings']['hg38'], array(
```
ifokkema, line 500: I don't feel comfortable having a certain genome build hardcoded into the source. How can we always pick the latest build? I think we can safely assume;
- The order of builds in `$_SETT['human_builds']` is from old to new.
- We won't add new builds unless VV supports them.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+
+    // When sending in a variant on transcript, VariantValidator only
+    //  returns the variant as mapped on that one transcript. If we are
+    //  on a VOT creation form, and there are multiple transcripts open,
+    //  we want each transcript to get a mapping. To get this, we then
+    //  need to call VariantValidator a second time using one of the
+    //  genomic variants as were returned using our first call.
+    if ($sType == 'VOT' && count($aTranscripts) > 1) {
+
+        $aMappedViaGB = (
```
ifokkema, line 497: This code is a bit long to be put in a ternary if(). How can this be made more readable?

src/ajax/check_hgvs_dialogue.php
ifokkema: So, basically, we should fill something in those fields (like `c.?`), and perhaps disable the `onchange()` because it won't do any good?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+        );
+
+        if (!isset($aMappedViaGB['data']['transcript_mappings'])) {
+            // If for any reason no new transcript mappings were given,
+            //  we cannot perform this extra step, and will thus miss
+            //  some information. We will inform the user.
+            update_dialogue('<br>Your variant could not be mapped to all transcripts due to unknown issues.');
+        }
+
+        $aMappedVariant['data']['transcript_mappings'] = $aMappedViaGB['data']['transcript_mappings'];
```
ifokkema, line 515: Since `update_dialogue()` doesn't exit, doesn't this cause notices?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+    // Returning the mapping for transcript, RNA and protein variants.
+    foreach ($aMappedVariant['data']['transcript_mappings'] as $sTranscript => $aTranscriptData) {
+        // Filling in the input fields.
+        print('
+        // Adding transcript info to the fields.        
+        var oTranscriptField = $("input").filter(function() { 
+            return $(this).data("id_ncbi") == "' . $sTranscript . '" 
+        });
+        
+        if (!oTranscriptField.prop("disabled")) {
```
ifokkema, line 543: We don't do this anymore, do we?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+            oTranscriptField.val("' . $aTranscriptData['DNA'] . '").attr("class", "acc");;
```
ifokkema, line 544: Earlier, `removeClass()` was used. Is that not used here to make sure other values are overwritten?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+            oTranscriptField.siblings("img:first").attr({src: "gfx/check.png", title: "Validated"}).show();
```
ifokkema, line 545: Upon first glance, I can only see that images carry a `.show()` command. I never see a `.hide()`. Are they ever hidden, or is their `src` always changed? If the latter is true, all `.show()` calls can be removed.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+
+        } else {
+            // Multiple possible variants were found. We will inform the user,
+            //  and concatenate the variants similarly to the example below:
+            // NC_1233456.1:g.1del + NC_123456.1:g.2_3del + NC_123456.1:g.4del =
+            //  NC_123456.1:g.1del^2_3del^4del.
+            update_dialogue('<br>There were multiple genomic variant predictions for build ' . $sGBID . '.');
+
+            $sMappedGenomicVariant =
+                preg_replace('/(.*:[a-z]\.).*/', '', $aMappedGenomicVariant[0]) .
```
ifokkema, line 572: This `preg_replace()` returns nothing.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+            $sMappedGenomicVariant = $aMappedGenomicVariant[0];
+
+        } else {
+            // Multiple possible variants were found. We will inform the user,
+            //  and concatenate the variants similarly to the example below:
+            // NC_1233456.1:g.1del + NC_123456.1:g.2_3del + NC_123456.1:g.4del =
+            //  NC_123456.1:g.1del^2_3del^4del.
+            update_dialogue('<br>There were multiple genomic variant predictions for build ' . $sGBID . '.');
+
+            $sMappedGenomicVariant =
```
ifokkema, line 571: This doesn't return "NC_123456.1:g.1del^2_3del^4del" as documented, but "1del^2_3del^4del".

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+                preg_replace('/(.*:[a-z]\.).*/', '', $aMappedGenomicVariant[0]) .
+                implode('^',
+                    array_map(function ($sFullVariant) {
+                        return preg_replace('/.*:[a-z]\./', '', $sFullVariant);
+                    }, $aMappedGenomicVariant)
+                );
+        }
+
+        // Removing the reference sequence.
+        $sMappedGenomicVariant = preg_replace('/.*:/', '', $sMappedGenomicVariant); // Fixme; Find a cleaner way of cutting off the reference sequence.
```
ifokkema, line 581: Personally, I always use `strstr()` with a true in argument 3. I guess the same could be used in the `preg_replace()` above. Since we know we have `refseq:prefix.variant` there, we could simply do a `strstr()` on a period (".") to merge the multiple variant descriptions.

src/class/object_genome_variants.php
```diff
@@ -268,6 +254,24 @@ function checkFields ($aData, $zData = false, $aOptions = array())
             }
         }
 
+        foreach ($aData as $sField => $sVariant) {
+            if (preg_match('/DNA/', $sField)) {
```
ifokkema, line 258: Use the much faster `strpos()` for this.

src/class/object_genome_variants.php
ifokkema: I'm not sure if this works the way intended. If the variant is not HGVS but does have an `IUNCERTAINPOSITIONS`, then it's accepted? Not sure if that's OK.
Also, I think we can make this a bit more efficient by first doing the `lovd_isHGVS()` check, and running `lovd_getVariantInfo()` only when `lovd_isHGVS()` returns false?

src/class/object_genome_variants.php
```diff
@@ -351,9 +355,15 @@ function getForm ()
...
-            (!$this->aColumns['VariantOnGenome/DNA']['description_form']? '' : '<BR>' . $this->aColumns['VariantOnGenome/DNA']['description_form']);
+        // Add genome build name and element data to VOG/DNA fields, and add in the link to the HGVS website.
+        $aActiveBuilds = $_DB->query('SELECT id, name, column_suffix FROM ' . TABLE_GENOME_BUILDS)->fetchAllGroupAssoc();
+        foreach ($aActiveBuilds as $sBuild => $aBuild) {
+            $sColumn = 'VariantOnGenome/DNA' . (!$aBuild['column_suffix']? '' : '/' . $aBuild['column_suffix']);
+            $this->aColumns[$sColumn]['description_form'] = '<B>Relative to ' . $aBuild['name'] . '.</B>' .
+                (!$this->aColumns[$sColumn]['description_form']? '' : '<BR>' . $this->aColumns[$sColumn]['description_form']);
+            $this->aColumns[$sColumn]['element_data'] = array('genome_build' => $sBuild);
+            $this->aColumns[$sColumn]['form_type'][0] = str_replace('(HGVS format)', '(<A href="http://varnomen.hgvs.org/recommendations/DNA" target="_blank">HGVS format</A>)', $this->aColumns[$sColumn]['form_type'][0]);
```
ifokkema, line 365: Minor thing; better refer to the site using `https` rather than `http`.

src/inc-js-variants.php
```diff
@@ -42,120 +42,52 @@
 define('AJAX_NO_AUTH', '8');
 define('AJAX_DATA_ERROR', '9');
 
-$_SETT = array('objectid_length' => array('transcripts' => 8));
+$_SETT = array(
+    'objectid_length' => array('transcripts' => 8),
+    'variant_validator' => array(
+        'genome_builds' => array('hg19', 'hg38')
```
ifokkema, line 48: Two requests;
- Better stick to the existing `$_SETT` structure, so the code in this file doesn't look different from the code in other files trying to reach the same information.
- Better add a comment to `inc-init.php` in the relevant place that we need to check this file if we ever change the builds. The same is already done for the `objectid_length` setting, so you can copy that format.

src/inc-js-variants.php
```diff
@@ -42,120 +42,52 @@
...
+    var oChromosome = $('select[name="chromosome"]');
+    var sChromosome = (
+        oChromosome.length ? // Yes=VOG form; No=VOT form.
+            oChromosome.val() :
+            $("td:contains('Chromosome')").last().filter(function () {
+                return $(this).html() === "Chromosome";
+            }).siblings().eq(1).html()
+    );
+    var sRefSeqInfo = (
+        sFieldName[0] !== 'V'? // Yes=RefSeq is a transcript; No=RefSeq is genomic.
```
ifokkema, line 76: I don't understand this logic? Don't all field names start with a 'V'?

src/inc-js-variants.php
```diff
@@ -42,120 +42,52 @@
...
+        $(this).data('id_ncbi') :
+        $(this).data('genomeBuild') + '-' + sChromosome
```
ifokkema, line 78: Why the choice for `-` instead of the more common `:`?

src/inc-js-variants.php
```diff
@@ -215,9 +149,16 @@ className: '',
         onmouseout: ''
     }).show();
 
-    if (oThisDNA.attr('name') == 'VariantOnGenome/DNA') {
+    if (oThisDNA.filter("[name^='VariantOnGenome/DNA']").size()) {
         // This function was called from the genomic variant, so build a list of genes and prepare the variant accordingly for mutalyzer.
-        var sVariantNotation = 'chr<?php echo $_GET['chromosome']; ?>:' + oThisDNA.val();
+        var sChromosome = (
+            oChromosome.length ? // Yes=VOG form; No=VOT form.
```
ifokkema, line 155: Where is this variable defined? Only in a different function, no?

src/inc-js-variants.php
```diff
@@ -239,14 +180,14 @@ className: '',
...
-                    if (oThisDNA.attr('name') == 'VariantOnGenome/DNA') {
+                    if (oThisDNA.filter("[name^='VariantOnGenome/DNA']").size()) {
                         // This function was called from the genomic variant, so fill in the return values from mutalyzer in the transcript DNA fields.
                         aVariants = sData.split(';');
                         var nVariants = aVariants.length;
                         for (i = 0; i < nVariants; i++) {
                             var aVariant = /^([A-Z]{2}_\d{6,9}\.\d{1,2}(?:\([A-Z0-9]+_v\d{3}\))?):([cn]\..+)$/.exec(aVariants[i]);
                             if (aVariant != null) {
-                                var oInput = $('#variantForm input[id_ncbi="' + aVariant[1] + '"]');
+                                var oInput = $('#variantForm input[data-id_ncbi="' + aVariant[1] + '"]');
```
ifokkema, line 190: Although this notation is easier, I think there's maybe a reason why the filters on the data fields normally work differently?

src/inc-js-variants.php
```diff
@@ -515,44 +456,47 @@ function lovd_highlightInput (oElement)
...
         if (oProteinVariants[0] != undefined) {
             // Add the buttons and images at the end of the protein description fields.
-            oProteinVariants.parent().append('&nbsp;&nbsp;<IMG src="gfx/trans.png" style="display:inline;" align="top" width="16" height="16">&nbsp;<BUTTON class="proteinChange" type="button" onclick="lovd_getProteinChange(this); return false;" style="display:none;">Predict</BUTTON>');
-        }
-
-        if (oTranscriptVariants.val() !== '') {
-            // Variant field already has content, check HGVS now because if we're on an edit form we
-            // want the buttons to be ready.
-            oTranscriptVariants.change();
+            oProteinVariants.parent().append('&nbsp;&nbsp;<IMG src="gfx/trans.png" style="display:none;" align="top" width="16" height="16">&nbsp;');
```
ifokkema, line 499: Ah, here I see one `"display:none"` that is removed everywhere else. If we remove this style, we can also get rid of all the `.show()` calls out there.

src/scripts/fix_variant_descriptions.php
```diff
@@ -549,7 +549,7 @@ function ($sVOT)
...
                                         || (strpos($aVOT['RNA'], 'spl') !== false && preg_match('/[0-9]+[+-][0-9]+/', $aVVVot['data']['DNA']))) {
                                         // We ignore small differences, where maybe the RNA has been verified.
-                                    } elseif (lovd_getVariantInfo(lovd_fixHGVS('c' . strstr($aVOT['RNA'], '.'), 'c'), '', true)
+                                    } elseif (lovd_isHGVS(lovd_fixHGVS('c' . strstr($aVOT['RNA'], '.'), 'c'))
```
ifokkema, line 552: This may not be correct; `''` is replaced by `false` here, which is something different.

src/scripts/fix_variant_descriptions.php
```diff
@@ -941,7 +941,7 @@ function ($sVOT)
...
```
ifokkema, line 944: This may not be correct; `''` is replaced by `false` here, which is something different.

ifokkema (changes requested):
I would like you to make some changes to this code - it should still be fresh in your memory, I hope :wink:

src/scripts/fix_variant_descriptions.php
loeswerkman: Yes, I discussed this with you when I changed this code: the `''` was incorrect, and it should have been `false` in the original code.

src/scripts/fix_variant_descriptions.php
loeswerkman: Yes, I discussed this with you when I changed this code: the `''` was incorrect, and it should have been `false` in the original code.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+    oGenomicVariants.val("").removeClass();
+    oGenomicVariants.siblings("img").attr({src: "gfx/trans.png"}).show();
+}
+');
+
+
+if (!$sVariant) {
+    // If the variant is empty, we can simply close the script.
+    exit;
+}
```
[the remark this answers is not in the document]
loeswerkman: Yes, this *does* mean that all fields are emptied if a user empties one of the fields. We discussed the best ways to reset all fields, and I think I remember you saying this was what we wanted. However, if that no longer seems to be the case, we could of course move this code up! Then the user would have to refresh the page to reset the input.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+
+
+
+// Preparing variables to allow easy and readable updates of the dialogue.
+
+// Preparing the steps.
+$sStepInitialChecks = 'statusChecks';
+$sStepMapping       = 'statusMapping';
```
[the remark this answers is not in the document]
loeswerkman: This is personal preference, I think. Personally I think working with variables instead of strings in these cases is easier to read and adjust. Constants would be a bit annoying since they could then be reached in other scripts and that would be unnecessarily confusing. If you prefer it to be a string, that is fine, too.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+
+// Preparing the images.
+$sImageNeutral = 'gfx/trans.png';
+$sImagePassed  = 'gfx/check.png';
+$sImageFailed  = 'gfx/cross.png';
+$sImageLoading = 'gfx/lovd_loading.gif';
```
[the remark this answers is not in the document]
loeswerkman: -> This is personal preference, I think. Personally I think working with variables instead of strings in these cases is easier to read and adjust. Constants would be a bit annoying since they could then be reached in other scripts and that would be unnecessarily confusing. If you prefer it to be a string, that is fine, too.

src/inc-js-variants.php
loeswerkman: Transcriptfields are formatted as follows: <ID>_VariantOnTranscript/DNA. Genomic fields are formatted like this: 'VariantOnGenome/DNA/<ID>'. So the 'V' only returns true for genomic fields. I do agree that this could be done in a much cleaner way.

src/ajax/check_hgvs_dialogue.php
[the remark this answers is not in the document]
ifokkema: It's definitely a personal preference, indeed.
Constants won't be accessible in other scripts, as other scripts would have to include this file to get those values; then the variables would also be included. If we'd change these values someday, then variables would be more useful indeed, but the values are only used in this script so that benefit is also limited. I think it's better to stick to the "common methods" used in LOVD everywhere else, and just use the strings and no variables.

src/ajax/check_hgvs_dialogue.php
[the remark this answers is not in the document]
ifokkema: Let's keep things "standard" and just use strings. Thanks!

src/ajax/check_hgvs_dialogue.php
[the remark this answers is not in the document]
ifokkema: After discussing this; better move this `exit` up, so that when no variant is passed, the DNA fields will *not* be emptied. When a variant has been passed, the current behavior of emptying all DNA fields is correct and should be kept.

src/ajax/check_hgvs_dialogue.php
ifokkema: After thinking about this some more, I guess removing the `onchange()` should only be done maybe if we're sure the transcript is the problem? Do we have examples of this situation?

ifokkema:
I also realized that we still need to modify the existing tests; I added this task to #562, but add this comment here as a reminder as well.

src/inc-js-variants.php
ifokkema: Ahh, right! I forgot about the transcript ID prefix.
We could drop the comment if we do a search for "Transcript" in the field name, then the code makes it obvious.

src/scripts/fix_variant_descriptions.php
ifokkema: You're right, I forgot about this and rushed to a conclusion in an attempt to finish my review before catching the train :sweat_smile: Sorry!

src/scripts/fix_variant_descriptions.php
ifokkema: You're right, I forgot about this and rushed to a conclusion in an attempt to finish my review before catching the train :sweat_smile: Sorry!

src/ajax/check_hgvs_dialogue.php
loeswerkman: Filling in empty fields seems right. Perhaps I could add the option to run the dialogue on one variant only? If this options would be turned on, the script would no longer reset all values, and would also not perform mapping (VariantValidator would only be run to validate the variant's own syntax). Adding this option might also prove useful when LOVD will allow non-human genome builds.

src/class/object_genome_variants.php
loeswerkman: You were right!! It has been fixed.

loeswerkman:
When 'something went wrong on our side', and the mapping is sent off but fails, the variant will be sent back to the input field with the reference sequence still attached to it. Find a clean way to fix this.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,603 @@
...
+/*******************************************************************************
+ *
+ * LEIDEN OPEN VARIATION DATABASE (LOVD)
+ *
+ * Created     : 2020-09-22
+ * Modified    : 2021-02-25
+ * For LOVD    : 3.0-27
+ *
+ * Copyright   : 2004-2021 Leiden University Medical Center; http://www.LUMC.nl/
+ * Programmer  : Ivo F.A.C. Fokkema <I.F.A.C.Fokkema@LUMC.nl>
```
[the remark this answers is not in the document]
ifokkema: De Created datum is nog steeds niet correct :wink:
De eerste commit van deze file is van 26 februari dit jaar, mogelijk is de juiste datum in februari of januari dit jaar?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -56,48 +64,23 @@
...
-$sButtonNo             = 'oButtonNo';
-$sButtonOKValid        = 'oButtonOKValid';
-$sButtonOKInvalid      = 'oButtonOKInvalid';
-$sButtonOKCouldBeValid = 'oButtonOKCouldBeValid';
 
 // Preparing the JS for the buttons.
 print('
 // Preparing the buttons.
-var ' . $sButtonYes . ' = {"Yes":function () {
+var ' . 'oButtonYes' . ' = {"Yes":function () {
```
ifokkema, line 83: Could you please remove the `' . '` string concatenations? :grimacing: 
(here and a few times below)
Thanks :sweat_smile:

src/ajax/check_hgvs_dialogue.php
```diff
@@ -252,7 +235,7 @@ function update_images_per_step($sStep, $sImage)
             // Good, we can propose a fix. If the user agrees with the fix,
             //  we can continue to the mapping.
             update_dialogue($sResponse . 'Did you mean \"' . $sFixedVariant . '\"?<br>',
-                $sButtonYes . ', ' . $sButtonNo);
+                'oButtonYes' . ', ' . 'oButtonNo');
```
ifokkema, line 241: This concatenation can also be cleaned up.

src/class/object_genome_variants.php
```diff
@@ -255,18 +255,20 @@ function checkFields ($aData, $zData = false, $aOptions = array())
         }
 
         foreach ($aData as $sField => $sVariant) {
-            if (preg_match('/DNA/', $sField)) {
+            if (strpos($sField, '/DNA/') !== false) {
```
ifokkema, line 258: This change is not the same, and I'm thinking this may fail sometimes.

ifokkema (changes requested):

src/ajax/check_hgvs_dialogue.php
```diff
@@ -575,17 +579,26 @@ function update_images_per_step($sStep, $sImage)
...
-            $(this).val("g.?");
-        }
-    });
-    $(\'#variantForm input[name*="VariantOnTranscript"]\').each(function(e){
-        if($(this).val()==="" && !$(this).attr("name").endsWith("Exon")) {
-            $(this).val("c.?");
+    $(\'#variantForm input[name*="VariantOn"]\').each(function(e){
+        sName = $(this).attr("name");
+        if ($(this).val() === ""
+            && !(sName.endsWith("Exon") || sName.endsWith("DBID") || sName.endsWith("Reference"))) {
```
ifokkema, line 594: Negative selection scares me - what if other fields are created that you may not have anticipated?

src/ajax/check_hgvs_dialogue.php
```diff
@@ -73,10 +69,10 @@
     // We know we got information on a GB. This is given through
     //  JS in the format of <genome build ID>-<chromosome>.
     $sType = 'VOG';
-    list($sGenomeBuildID, $sChromosome) = explode('-', $sRefSeqInfo);
+    list($sGenomeBuildID, $sChromosome) = explode('-', urldecode($_REQUEST['refSeqInfo']));
```
ifokkema, line 72: Last one about `urldecode()` :wink:

src/ajax/check_hgvs_dialogue.php
```diff
@@ -128,15 +124,15 @@
     // We will fill in this fixed variant, close the dialogue,
     //  and perform a new call to this script by activating
     //  the onChange.
-    var oInput = $(\'input[name="' . $sFieldName . '"]\');
-    oInput.val("' . addslashes(lovd_fixHGVS($sVariant)) . '");
+    var oInput = $(\'input[name="' . htmlspecialchars($sFieldName) . '"]\');
+    oInput.val("' . htmlspecialchars(lovd_fixHGVS($_REQUEST['var'])) . '");
```
ifokkema, line 128: `addslashes()` may have been enough (it escapes double quote also), although `htmlspecialchars()` is more thorough. I hope this won't cause issues with substitutions; was this checked? (just to be sure)

src/ajax/check_hgvs_dialogue.php
```diff
@@ -128,15 +124,15 @@
     // We will fill in this fixed variant, close the dialogue,
     //  and perform a new call to this script by activating
     //  the onChange.
-    var oInput = $(\'input[name="' . $sFieldName . '"]\');
-    oInput.val("' . addslashes(lovd_fixHGVS($sVariant)) . '");
+    var oInput = $(\'input[name="' . htmlspecialchars($sFieldName) . '"]\');
```
ifokkema, line 127: Just like with `$sVariant` earlier, it's probably best to leave this as its `$_REQUEST` counterpart or clean it beforehand. This way, `$sFieldName` is not mistaken for a cleaned variable.

ifokkema (changes requested):
There's some small stuff left, and I also saw some open comments (e.g., about `ACTION`).

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,610 @@
...
+    // The '-' serves as a communication tool; it tells us that
+    //  the given input was a GB + chromosome. When no '-' is
+    //  found, we know that the input was the reference sequence
+    //  of a transcript.
+    $sType = 'VOT';
+    $sReferenceSequence = urldecode($_REQUEST['refSeqInfo']);
+    global $_DB;
+    $bRefSeqIsSupportedByVV = (
+        'hg' == substr($_DB->query('SELECT id FROM ' . TABLE_GENOME_BUILDS . ' LIMIT 1')->fetchColumn(), 0, 2)
+    );
```
ifokkema, line 66: I only noticed this today;
- Global isn't necessary because you're not in a function,
- If you want to do a `substr()` on a result from MySQL, you'd better just let the database handle it (see MySQL's `LEFT()`).

src/ajax/check_hgvs_dialogue.php
```diff
@@ -35,35 +35,31 @@
...
+$aTranscripts = (empty($_REQUEST['transcripts'])? array() : explode('|', urldecode($_REQUEST['transcripts'])));
 
 // Retrieving the name of the input field.
-$sFieldName   = htmlspecialchars($_REQUEST['fieldName']);
+$sFieldName = urldecode($_REQUEST['fieldName']);
 
 
 
-if (!$sVariant) {
+if (!($_REQUEST['var'])) {
```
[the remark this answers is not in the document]
loeswerkman: Whoops, it was!

src/ajax/check_hgvs_dialogue.php
loeswerkman: Fixed!

src/ajax/check_hgvs_dialogue.php
```diff
@@ -35,35 +35,31 @@
...
 }
 
 
 
-// Retrieving information on the reference sequence from the URL.
-$sRefSeqInfo  = htmlspecialchars($_REQUEST['refSeqInfo']);
-
 // Retrieve the reference sequence from the info given through the URL.
-if (strpos($sRefSeqInfo, '-') === false) {
+if (strpos($_REQUEST['refSeqInfo'], '-') === false) {
```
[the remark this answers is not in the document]
loeswerkman: done!

src/ajax/check_hgvs_dialogue.php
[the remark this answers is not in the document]
loeswerkman: Noted! Clear explanation and the fix has been implemented :)

ifokkema:
Note to self; when going through this code, check how the form handles `g.(1234_?)_(?_2345)del`; the fix that we provide is also not correct (`g.(1234_2345)del`, suffix required). Perhaps use part of the code used in the API and the check HGVS interface that also looks at the errors thrown at the fixed variant.

PR #594, opened 2022-04-04
Improve/api variants

### LINKED ISSUE #589: Dynamically find and show all active GBs in the LOVD API
This issue aims to remove the hard-coded approach of the original `api.php` code. Instead, we want to find all active GBs in the database and, based on that, show the information of each variant on this build.

loeswerkman:
This branch updates the code that allows variants to be viewed using LOVD APIs. The code will be refactored and extended to dynamically handle different active genome builds.

Closes #589. Related to #550.


src/class/api.ga4gh.php
```diff
@@ -42,6 +43,8 @@ class LOVD_API_GA4GH
     private $API;                     // The API object.
     private $aURLElements = array();  // The current URL broken in parts.
     private $aFilters = array();      // Filters active on the data.
+    private $sActiveGB;
```
loeswerkman, line 46: Translated feedback from email @ifokkema:

>Use `$sPrimaryGB` or `$sFirstGB` to avoid confusion with $aActiveGBs.

src/class/api.ga4gh.php
```diff
@@ -211,6 +214,11 @@ function __construct (&$oAPI)
         // To not duplicate code.
         $this->aValueMappings['effect'] = $_SETT['var_effect_api'];
 
+        // Retrieve default genome build.
+        global $_DB;
+        $this->aActiveGBs = $_DB->query('SELECT id, column_suffix FROM ' . TABLE_GENOME_BUILDS)->fetchAllGroupAssoc();
+        $this->sActiveGB = array_keys($this->aActiveGBs)[0];
```
loeswerkman, line 220: Translated feedback from email @ifokkema:

>This is not available in PHP<5.4.0. Use `array_key_first()` instead.

src/class/api.ga4gh.php
```diff
@@ -244,6 +252,37 @@ private function addComment ($aComments, $Text, $sTerm = '')
...
+                        'source' => 'genbank',
+                        'accession' => $_SETT['human_builds'][$sGBID]['ncbi_sequences'][$sChr],
+                    ),
+                    'name' => array(
+                        'scheme' => 'HGVS',
+                        'value' => $zData['DNA' . $sGBID],
+                    ),
+                );
+            },
+            array_filter(array_keys($this->aActiveGBs)), function($sGBID) use ($sBuild) {
```
loeswerkman, line 276: Translated feedback from email @ifokkema:

>Use `array_diff()` instead of `array_filter()`.

src/class/api.ga4gh.php
```diff
@@ -668,21 +705,19 @@ private function showTableData ($sTableName)
...
-        global $_CONF, $_SETT;
+        global $_SETT;
 
         if ($sTableName == 'variants') {
             list(, $sBuild, $sChr, $sPosition) = array_pad($aPage, 4, '1');
 
-            if ($sBuild != $_CONF['refseq_build']) {
-                // We don't support this yet, because we can't use an index on a
-                //  search on the other genome build. Wait for LOVD 3.1.
+            if (!in_array($sBuild, array_keys($this->aActiveGBs))) {
```
loeswerkman, line 713: Translated feedback from email @ifokkema:

>Use `isset()` instead of combining `in_array()` and `array_keys()`.

src/class/api.ga4gh.php
ifokkema: Yeah, so I see I didn't pay attention there - `array_key_first()` is even newer (PHP >= 7.3.0) so is not a good replacement. `key()` will probably do the trick, but should be tested. If it doesn't work, `current()` over the `array_keys()` should always work.

src/class/api.ga4gh.php
```diff
@@ -777,14 +812,30 @@ private function showVariantDataPage ($sBuild, $sChr, $sPosition)
...
             'VariantOnGenome/Reference',
             'VariantOnGenome/Remarks',
         ));
+        $sGenomeBuildQ = '';
+        foreach ($this->aActiveGBs as $sGBID => $sGBSuffix) {
+            // We will prepare to get the right data for each active
+            //  genome build.
+            $sPreparedSuffix = !$sGBSuffix ? '' : '/' . $sGBSuffix;
+
+            if ($sGBID != $sBuild) {
```
loeswerkman, line 828: Copy of feedback from email @ifokkema:

>Zoals eerder genoemd is het flink minder code als je die `if()` weg haalt; dan hoeft `$aRequiredCols` niet vooraf de VOG/DNA kolom te bevatten en de query `$sQ` hoeft dan ook niet al het DNA veld te bevatten. Je zou dan ook nog de selectie van de `vog.position_g_start` en `vog.position_g_end` velden "mooier" kunnen maken door ze binnen de aangegeven `foreach()` toe te laten voegen, gebruik makend van `$sPreparedSuffix` (je hebt voor die twee velden dan wel een `if()` nodig).

src/class/api.ga4gh.php
loeswerkman: `key()` did the trick :)

ifokkema: Excellent!

src/api.php
```diff
@@ -600,17 +644,20 @@
...
             'position_mRNA' => $zData['_position_mRNA'],
-            'position_genomic' => array(
-                $_CONF['refseq_build'] => $zData['position_genomic'],
-            ),
-            'variant_genomic' => array(
-                $_CONF['refseq_build'] => 'chr' . $sChromosome . ':' .
-                    $zData['VariantOnGenome/DNA'],
-            ),
+            'position_genomic' => $aGenomicPositions,
+            'variant_genomic' => $aGenomicVariants,
```
ifokkema, line 660: Note that in the original code, `position_genomic` and `variant_genomic` were only returned as arrays if we asked for JSON output. Which genome build is now selected when we don't ask for JSON output? Is that still always the "correct" one? (I don't think so)

src/class/api.ga4gh.php
```diff
@@ -211,6 +214,11 @@ function __construct (&$oAPI)
         // To not duplicate code.
         $this->aValueMappings['effect'] = $_SETT['var_effect_api'];
 
+        // Retrieve default genome build.
+        global $_DB;
```
ifokkema, line 218: Globals can be defined at the top of the function/method.

src/class/api.ga4gh.php
```diff
@@ -668,21 +716,19 @@ private function showTableData ($sTableName)
...
-                if (!isset($_SETT['human_builds'][$sBuild])) {
-                    $this->API->aResponse = array('errors' => array('title' => 'Unrecognized genome build.'));
-                } else {
-                    $this->API->aResponse = array('errors' => array('title' => 'Unsupported genome build.'));
-                }
-                $this->API->aResponse['errors']['detail'] = 'We can not use genome build ' . $sBuild . '. Please choose from: \'' . $_CONF['refseq_build'] . '\'.';
+                $this->API->aResponse = array('errors' => array('title' =>
+                    (isset($_SETT['human_builds'][$sBuild])? 'Unsupported' : 'Unrecognized')  . ' genome build.')
+                );
+                $this->API->aResponse['errors']['detail'] = 'We cannot use genome build ' . $sBuild . '. Please choose from: \'' . implode('|', array_keys($this->aActiveGBs)) . '\'.';
```
ifokkema, line 731: This will look a bit weird;
```
Please choose from: 'hg19'.
```
looks OK, but 
```
Please choose from: 'hg19|hg38'.
```
Looks a bit weird. Probably `'hg19', 'hg38'` would be more logical for humans.

src/class/api.ga4gh.php
```diff
@@ -211,6 +214,11 @@ function __construct (&$oAPI)
         // To not duplicate code.
         $this->aValueMappings['effect'] = $_SETT['var_effect_api'];
 
+        // Retrieve default genome build.
+        global $_DB;
+        $this->aActiveGBs = $_DB->query('SELECT id, column_suffix FROM ' . TABLE_GENOME_BUILDS)->fetchAllCombine();
```
ifokkema, line 219: I think that also here, we can fix a lot of suffix issues if we'd let the database concatenate the suffixes into the column names for us once, and we can just retrieve the required fields from this `SELECT` result.

src/class/api.ga4gh.php
```diff
@@ -862,10 +932,9 @@ private function showVariantDataPage ($sBuild, $sChr, $sPosition)
                  GROUP_CONCAT(DISTINCT t.geneid ORDER BY t.geneid SEPARATOR ";") AS genes,
                  GROUP_CONCAT(DISTINCT
                    IFNULL(i.id,
-                     CONCAT(vog.id, "||", IFNULL(uc.default_license, ""), "||"' .
-            (!$bDNA38? '' : ',
-                       IFNULL(vog.`VariantOnGenome/DNA/hg38`, "")') . ', "||",
-                       vog.effectid, "||"' .
+                     CONCAT(vog.id, "||", IFNULL(uc.default_license, ""), "||", vog.effectid, "||",
+                        IFNULL(vog.`VariantOnGenome/DNA' . (!$this->aActiveGBs[$sBuild]? '' : '/' . $this->aActiveGBs[$sBuild]) . '`, ""), "||"' .
+
```
ifokkema, line 937: I don't think this is the same functionality. The query used to fetch an alias if it was available. Now it seems to just fetch the selected build's DNA field again.

src/class/api.ga4gh.php
```diff
@@ -1286,9 +1334,7 @@ function ($aPathogenicity) {
                           vog.id, "||",
                           vog.allele, "||",
                           vog.chromosome, "||",
-                          vog.`VariantOnGenome/DNA`, "||"' .
-                    (!$bDNA38? '' : ',
-                       IFNULL(vog.`VariantOnGenome/DNA/hg38`, "")') . ', "||",
+                          vog.`VariantOnGenome/DNA' . (!$this->aActiveGBs[$sBuild]? '' : '/' . $this->aActiveGBs[$sBuild]) . '`, "||",
```
ifokkema, line 1337: Functionality seems to be lost here?
Did you test if the output is the same (or similar) than the output of LOVD 3.0 for a certain variant?

ifokkema (changes requested):
I'm unsure about a few changes.
I also think we can make the code more readable by extending the query that fetches the available genome builds. If you're not sure about my suggestion on that part, let me know!

src/api.php
```diff
@@ -239,7 +240,25 @@
...
-        // First build query.
+
+        // Prepare the SQL that will look for all genomic data.
+        // We will go through all active builds and prepare a query
+        //  that retrieves the information from each build individually.
+        $sPreparedGenomicQ = '';
+        foreach ($aActiveGBs as $sGBID => $sGBSuffix) {
+            $sUnderscoreSuffix = (!$sGBSuffix? '' : '_' . $sGBSuffix);
+            $sPreparedGenomicQ .= '
+                vog.chromosome, vog.`VariantOnGenome/DNA' . (!$sGBSuffix? '' : '/' . $sGBSuffix) . '`,
```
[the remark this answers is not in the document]
ifokkema: This suggestion should maybe be ignored, considering my suggestion elsewhere to do this differently.

src/api.php
```diff
@@ -322,30 +333,63 @@
...
                                 if (empty($aRegs[3])) {
                                     // No range. Absolute location.
                                     $aRegs[4] = $aRegs[2];
                                 }
                                 // Very important in genomic positions: genes on antisense will have positions like g.5678_1234 in the database!!!
-                                $nMin = min($aRegs[2], $aRegs[4]);
-                                $nMax = max($aRegs[2], $aRegs[4]);
+                                $nMin = (int) min($aRegs[2], $aRegs[4]);
+                                $nMax = (int) max($aRegs[2], $aRegs[4]);
+                                $sUnderscorePlusSuffix = (!$aActiveGBs[$sGB]? '' : '_' . $aActiveGBs[$sGB]);
```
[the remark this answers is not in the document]
loeswerkman: Hm, is it true you mean to build `$aActiveGBs` with a query like this one:
`SELECT id, if(column_suffix = "", "", CONCAT("/", column_suffix)) as suffix FROM lovd_genome_builds`?

I wanted to do this myself as well, but the problem I found was that the suffix cannot always be concatenated to the same symbol; so it is `/` for the DNA field, but `_` for the position fields. However, I could imagine that maybe something like:
`SELECT id, if(column_suffix = "", "", CONCAT("/", column_suffix)) as slashSuffix, if(column_suffix = "", "", CONCAT("_", column_suffix)) as underscoreSuffix FROM lovd_genome_builds;` to be an improvement!

Maybe I am going in the wrong direction, but this second option seems good to me!
Let me know what you think :)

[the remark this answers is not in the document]
loeswerkman: Oops, I had missed this example: 
> E.g., instead of `vog.position_g_start' . $sUnderscorePlusSuffix` we can do `$aActiveGBs[$sGB]['position_start']`.

This is fine as well; I will make it work!

src/api.php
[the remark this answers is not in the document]
loeswerkman: Done :)

src/api.php
```diff
@@ -322,30 +333,63 @@
...
+                            && (preg_match('/^(c)\.([*-]?[0-9]+([+-][du]?[0-9]+)?)(_([*-]?[0-9]+([+-][du]?[0-9]+)?))?$/', $_GET['search_' . $sField], $aRegs)
+                                // $aRegs:     1    2           3                  4 5           6 Fixme; This code does not support m. and n. DNA.
+                                || preg_match('/^(g)\.([0-9]+)(_([0-9]+))?$/', substr($_GET['search_position'], strpos($_GET['search_position'], ':')), $aRegs))
+                                // $aRegs:        1    2       3 4
+                        ) {
                             if ($aRegs[1] == 'g') {
+                                // The user is looking for a specific genomic position.
+                                if (strpos($_GET['search_position'], ':') !== false) {
+                                    // If a ':' was found, the user specified a genome build.
+                                    list($sGB, $searchPosition) = explode(':', $_GET['search_position']);
```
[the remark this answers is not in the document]
loeswerkman: Done with strstr :)

src/api.php
```diff
@@ -322,30 +333,63 @@
...
+                                        //  database. We will send an error.
+                                        header('HTTP/1.0 404 Not Found');
+                                        die('The given genome build was not found in the database.');
+                                    }
+
+                                } else {
+                                    // The user did not specify a specific genome build.
+                                    if (count($aActiveGBs) == 1) {
+                                        // If there is only one build, we can simply use that one.
+                                        $sGB = array_keys($aActiveGBs)[0]; // Fixme; Can this be done? Function()[]?
```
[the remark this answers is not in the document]
loeswerkman: Whoopsie, done with key()!

src/api.php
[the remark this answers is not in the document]
loeswerkman: Suffix comments are all processed to indeed work with `$aActiveGBs[$sGB][<field>]`.

src/class/api.ga4gh.php
loeswerkman: Moved :)

src/class/api.ga4gh.php
loeswerkman: Done as recommended!

src/class/api.ga4gh.php
loeswerkman: Yep you were very right; this has been done

src/class/api.ga4gh.php
loeswerkman: You are right! This was me being a bit clumsy and not removing this while it should have been removed! Later on in the code (read: around line 1110), this big concatenated string is exploded and expects for the item on line 936 to be removed. Because the items after this line were empty in all situations where I tested this, it did not come up!
Thank you for noticing and it has been fixed!

src/class/api.ga4gh.php
loeswerkman: Oopsie yes, if you refer to forgetting the `IFNULL()`,  it has now been added, tested and looks good!

src/api.php
```diff
@@ -309,7 +320,7 @@
                  LEFT OUTER JOIN ' . TABLE_INDIVIDUALS . ' AS i ON (s.individualid = i.id AND i.statusid >= ' . STATUS_MARKED . ')
                  LEFT OUTER JOIN ' . TABLE_USERS . ' AS uc ON (vog.created_by = uc.id)
                  LEFT OUTER JOIN ' . TABLE_USERS . ' AS uo ON (vog.owned_by = uo.id)
-               WHERE ' . ($nRefSeqID? 't.id = ' . $nRefSeqID : 't.geneid = ' . $_DB->quote($sSymbol)) . ' AND vog.statusid >= ' . STATUS_MARKED;
+               WHERE vog.chromosome = ' . $sChromosome . ' AND ' . ($nRefSeqID? 't.id = ' . $nRefSeqID : 't.geneid = ' . $_DB->quote($sSymbol)) . ' AND vog.statusid >= ' . STATUS_MARKED;
```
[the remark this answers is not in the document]
loeswerkman: Done !

src/class/api.ga4gh.php
ifokkema: I assume that's what I meant indeed. Thanks!

src/api.php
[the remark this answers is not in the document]
ifokkema: Excellent, thanks!

src/api.php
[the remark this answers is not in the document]
ifokkema: Perfect!

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,614 @@
...
+ *************/
+
+define('ROOT_PATH', '../');
+require ROOT_PATH . 'inc-init.php';
+require ROOT_PATH . 'inc-lib-variants.php';
+header('Content-type: text/javascript; charset=UTF-8');
+
+
+// Retrieving the variant and the transcripts to map to.
+$sVariant     = $_REQUEST['var'];
```
ifokkema, line 39: I can not properly follow this variable everywhere, to see if it's sanitized everywhere. Better not rename it, or sanitize it here. See also the two comments below.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -0,0 +1,614 @@
...
+$aTranscripts = (empty($_REQUEST['transcripts'])? array() : explode('|', $_REQUEST['transcripts']));
```
ifokkema, line 40: This variable hasn't been sanitized and is used unsanitized (danger!). Better sanitize it here or don't rename it. See also the comment below.

src/class/api.ga4gh.php
```diff
@@ -1,1957 +1,2014 @@
...
-        if (!$sChr) {
-            unset($aOutput['pagination']['next_page_url']);
-        }
-
-        $this->API->aResponse = $aOutput;
-        return true;
-    }
-}
-?>
+<?php
```
ifokkema: Please fix the line endings; it's CRLF again!

ifokkema (changes requested):
I haven't looked in much detail, but some changes are needed.

PR #597, opened 2022-04-08
Add/remove indices when adding/removing GBs

loeswerkman:
From now on, when a new genome build is added, a new index will be built for the new position fields, and this index will be removed again when the build is removed.

Related to #550.

### DIFF
--- a/src/genome_builds.php
+++ b/src/genome_builds.php
@@ -167,7 +167,14 @@ if (PATH_COUNT == 1 && ACTION == 'add') {
                 }
 
                 if ($bToAdd) {
-                    $_DB->query(rtrim($sSQL, ','));
+                    $_DB->query(
+                        ($sTable != TABLE_VARIANTS?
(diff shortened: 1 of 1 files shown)


src/genome_builds.php
```diff
@@ -167,7 +167,14 @@
...
 
                 if ($bToAdd) {
-                    $_DB->query(rtrim($sSQL, ','));
+                    $_DB->query(
+                        ($sTable != TABLE_VARIANTS?
+                            rtrim($sSQL, ',') :
+                            $sSQL .
+                            ' ADD INDEX (chromosome, position_g_start_' . $_POST['column_suffix'] .
+                            ', position_g_end_' . $_POST['column_suffix'] . ')'
+                        )
```
ifokkema, line 176: I think it's a bit more readable to make this into a "normal" `if()`, run before the `$_DB->query()` command.

ifokkema (approved):

PR #601, opened 2022-05-25
Feat/liftover existing variants

### LINKED ISSUE #600: Automatically lift over existing variants to all active GBs
Whenever a new genome buid is added, we should automatically perform lift overs from all existing variants to the new build the moment the columns of the new build are added to the `Variants` table. This can be done in `ajax/mapping_variants.php` using VariantValidator.

Related to #550.

loeswerkman:
This code will identify existing variants that are not described on all active GBs, and add missing descriptions wherever possible by performing lift overs using VariantValidator. 

Closes #600, related to #550.

### DIFF
--- a/src/ajax/map_variants.php
+++ b/src/ajax/map_variants.php
@@ -245,7 +245,7 @@ if (count($aActiveGBs) > 1) {
             ' LIMIT ' . (int) $nMaxVariants,
             array($sChr)
         )->fetchAllColumn();
-        $aVariantUpdates = array_merge($aVariantUpdates, $aVariantIDs);
+        $aVariantUpdates = array_fill_keys($aVariantIDs, true);
 
         // These variant are now in progress, which we will indicate by mapping flags.
         $_DB->query(


src/ajax/map_variants.php
```diff
@@ -198,6 +198,166 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+
+                    // Great! All checks have passed and the new description
+                    //  can be sent to the database.
+                    $aNewVariant = lovd_getVariantInfo($aVVResponse[$sBuild]);
+                    $_DB->query(
+                        'UPDATE ' . TABLE_VARIANTS .
+                        ' SET ' . $aGBColumns['DNA'] . ' = ?' .
+                        ', ' . $aGBColumns['position_start'] . ' = ?' .
+                        ', ' . $aGBColumns['position_end'] . ' = ?' .
+                        ', mapping_flags = mapping_flags | ' . MAPPING_DONE .
```
loeswerkman, line 320: Can we really add `MAPPING_DONE`? The lift overs are done, but what about the transcripts?

src/ajax/map_variants.php
```diff
@@ -309,9 +309,38 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
                         continue;
                     }
 
-                    // Great! All checks have passed and the new description
-                    //  can be sent to the database.
-                    $aNewVariant = lovd_getVariantInfo($aVVResponse['data'][$sBuild]);
+                    // Great! All checks have passed. Now all that is left to do is to
+                    //  ensure the variants are correctly concatenated in case multiple
+                    //  descriptions were found on the build in question.
+                    if (is_string($aVVResponse['data'][$sBuild])) {
```
loeswerkman, line 315: This whole thing (line 315-338) is also performed in `ajax/check_hgvs_dialogue.php` in the branch `improve/variantCreationForm`. Perhaps we could unify the code by building a function for both cases?

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+// Performing lift-overs so variants are described as relative to as many GBs as possible.
+$aActiveGBs = $_DB->query('
+        SELECT id,
+            CONCAT("VariantOnGenome/DNA", if(column_suffix = "", "", "/"), column_suffix) as "DNA",
+            CONCAT("position_g_start", if(column_suffix = "", "", "_"), column_suffix) as "position_start",
+            CONCAT("position_g_end", if(column_suffix = "", "", "_"), column_suffix) as "position_end"
+        FROM ' . TABLE_GENOME_BUILDS
+)->fetchAllGroupAssoc();
+
+if (count($aActiveGBs) < 1) {
```
ifokkema, line 211: This `if()` can never be true. How has the code been tested? How many builds were active?

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+    // Lift overs can only be performed if more than one GB is active.
+
+    foreach ($aActiveGBs as $sBuild => $aGBColumns) {
+        $aChr = $_DB->query(
+            'SELECT DISTINCT chromosome FROM ' . TABLE_VARIANTS .
+            ' WHERE ' . $aGBColumns['DNA'] . '= ""' . // DNA field of this GB is empty
```
ifokkema, line 217: Or, this field could be `NULL`!

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+            '    AND mapping_flags & ' . MAPPING_ALLOW . // Mapping is allowed
```
ifokkema, line 218: This line should be removed; also variants that don't allow mapping (VOG to VOT) should allow liftover (VOG to VOG) because we (LOVD) simply enforce it.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+            '    AND NOT mapping_flags & ' . (MAPPING_NOT_RECOGNIZED | MAPPING_DONE) // Mapping is possible and necessary
```
ifokkema, line 219: `MAPPING_DONE` should also be removed. However, you can remove `MAPPING_IN_PROGRESS`.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+        if (!empty($aChr)) {
+            // There is more than one chromosome with missing variant description for this build.
+
+            // We pick a random chromosome to minimize the chances of multiple active users
+            //  activating lift overs on the same chromosome.
+            $sChr = array_rand($aChr);
+
+            $aVariantIDs = $_DB->query(
+                'SELECT id FROM ' . TABLE_VARIANTS .
+                ' WHERE chromosome = ' . $sChr . // Only taking our picked chromosome
```
ifokkema, line 231: `$sChr` can be a string! That would cause this query to crash. Always use the SQL placeholders ("?") everywhere unless it's really not possible.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+                '    AND ' . $aGBColumns['DNA'] . '= ""' . // DNA field of this GB is empty
```
ifokkema, line 232: Was this query tested? I'm pretty sure this should cause an SQL error.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+                '    AND mapping_flags & ' . MAPPING_ALLOW . // Mapping  is allowed
+                '    AND NOT mapping_flags & ' . (MAPPING_NOT_RECOGNIZED | MAPPING_DONE) . // Mapping is possible and necessary
```
ifokkema, line 234: See also above, `MAPPING_ALLOW` shouldn't be considered, and `MAPPING_DONE` should be `MAPPING_IN_PROGRESS`.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+
+            // We will loop through all variants that are missing DNA descriptions.
+            foreach ($aVariantIDs as $sVariantID) {
+
+                // This variant is now in progress, which we will indicate by mapping flags.
+                $_DB->query(
+                    'UPDATE ' . TABLE_VARIANTS .
+                    ' SET mapping_flags = mapping_flags | ' . MAPPING_IN_PROGRESS .
+                    ' WHERE id = ?', array($sVariantID)
+                );
```
ifokkema, line 246: Applying this only inside the loop means that you can select a variant and only seconds later mark it as in progress. The mark should be stored immediately after selection, and whether that update was successful (whether you successfully "locked" the variants) should be checked.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+
+                list($bMappingTryAgain, $bMappingSuccessful) = array(false, false);
```
ifokkema, line 248: Easier is:
```php
$bMappingTryAgain = $bMappingSuccessful = false;
```
It doesn't require building an array and then decomposing it again.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+                'SELECT id FROM ' . TABLE_VARIANTS .
+                ' WHERE chromosome = ' . $sChr . // Only taking our picked chromosome
+                '    AND ' . $aGBColumns['DNA'] . '= ""' . // DNA field of this GB is empty
+                '    AND mapping_flags & ' . MAPPING_ALLOW . // Mapping  is allowed
+                '    AND NOT mapping_flags & ' . (MAPPING_NOT_RECOGNIZED | MAPPING_DONE) . // Mapping is possible and necessary
+                ' LIMIT ' . $nMaxVariants
+            )->fetchColumn();
+
+            // We will loop through all variants that are missing DNA descriptions.
+            foreach ($aVariantIDs as $sVariantID) {
```
ifokkema, line 239: These IDs are numeric, so `$nVariantID` would fit better.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+                    }
+
+                    $sVariant = $_DB->query('SELECT VariantOnGenome/DNA FROM ' . TABLE_VARIANTS .
+                        ' WHERE id = ' . $sVariantID)->fetchColumn();
+                    if ($sVariant == '') {
+                        // Variant description is empty; no chance of lifting over from this build.
+                        continue;
+                    }
+
+                    $aVariant = lovd_getVariantInfo($sVariantID);
```
ifokkema, line 265: I think you meant to put the DNA description there?

src/ajax/map_variants.php
loeswerkman: Resolved! We do not want to add `MAPPING_DONE`.

src/ajax/map_variants.php
ifokkema: In that case, that should then be in the VV library. I'm not really sure yet if maybe this entire problem should be solved inside of the VV library instead of outside of it.

src/ajax/map_variants.php
ifokkema: That is correct indeed, the flag should not be added.

src/ajax/map_variants.php
```diff
@@ -198,6 +198,195 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
...
+
+                // Now we will loop through all alternative builds to try lift overs
+                //  from all possible references.
+                foreach ($aActiveGBs as $sSourceBuild => $aSourceGBColumns) {
+                    if ($sSourceBuild == $sBuild) {
+                        // We need to lift over TO $sBuild, so not FROM $sBuild.
+                        continue;
+                    }
+
+                    $sVariant = $_DB->query('SELECT VariantOnGenome/DNA FROM ' . TABLE_VARIANTS .
```
ifokkema, line 258: Would it perhaps make sense to think about handling multiple empty fields? E.g., whenever we'd have three active builds and only one is filled in, it wouldn't make sense to fill the two empty fields one by one.

ifokkema (commented):
I realized too late you were still working on this - here are some comments already!

src/ajax/map_variants.php
loeswerkman: I think it would be nice to do so inside the library!

src/ajax/map_variants.php
loeswerkman: Oops ':)

src/ajax/map_variants.php
loeswerkman: Ah I misunderstood! Fixed :)

src/ajax/map_variants.php
loeswerkman: Yep on it:)

src/ajax/map_variants.php
loeswerkman: Done

src/ajax/map_variants.php
loeswerkman: Oh great, thank you!

src/ajax/map_variants.php
loeswerkman: Ah oopsie, you're right :)

src/ajax/map_variants.php
loeswerkman: Oh right, this is now taken into account!

src/ajax/map_variants.php
loeswerkman: Done

src/ajax/map_variants.php
```diff
@@ -228,11 +227,11 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
 
             $aVariantIDs = $_DB->query(
                 'SELECT id FROM ' . TABLE_VARIANTS .
-                ' WHERE chromosome = ' . $sChr . // Only taking our picked chromosome
-                '    AND ' . $aGBColumns['DNA'] . '= ""' . // DNA field of this GB is empty
-                '    AND mapping_flags & ' . MAPPING_ALLOW . // Mapping  is allowed
+                ' WHERE chromosome = ?' . // Only taking our picked chromosome
+                '    AND (' . $aGBColumns['DNA'] . ' = "" OR ISNULL(' . $aGBColumns['DNA'] . '))' . // DNA field is empty
```
ifokkema, line 231: In LOVD we always use `column IS NULL` instead of `ISNULL(column)`.
Also, note that `$aGBColumns['DNA']` will need some quoting as MySQL doesn't like column names with slashes. Check other queries to see how that's done.

src/ajax/map_variants.php
```diff
@@ -255,14 +254,14 @@ function lovd_mapVariantToTranscripts (&$aVariant, $aTranscripts)
                         continue;
                     }
 
-                    $sVariant = $_DB->query('SELECT VariantOnGenome/DNA FROM ' . TABLE_VARIANTS .
-                        ' WHERE id = ' . $sVariantID)->fetchColumn();
+                    $sVariant = $_DB->query("SELECT '{$aSourceGBColumns['DNA']}' FROM " . TABLE_VARIANTS .
```
ifokkema, line 257: Columns should be quoted in a different way. Check other queries to see how.

src/ajax/map_variants.php
loeswerkman: Noted and implemented!

src/ajax/map_variants.php
loeswerkman: Yes, definitely! Working on it :)

loeswerkman:
The code currently seems to work as I want it to. There is one problem still: the `MAPPING_IN_PROGRESS` flag that should be removed by the `lovd_updateVariantsOnExit()` shutdown function, is in fact not removed after the process. As of yet I do not understand why it is not. I will look into this! Also, I will continue to test this code before the actual branch should be merged. However, for now the code should be good for a review.

PR #603, opened 2022-06-09
Feat/force validation

loeswerkman:
In this branch, the `$_SESSION['VV']['validated_variants']` array is added and used to store all validated variants from the submission dialogue. This array is then used in `checkFields()` at `POST` to ensure that all variants in the submission form were indeed validated. 

Related to #588

### DIFF
--- a/src/ajax/check_hgvs_dialogue.php
+++ b/src/ajax/check_hgvs_dialogue.php
@@ -247,7 +247,7 @@ if ($_REQUEST['action'] == 'check') {
             'oButtonOKCouldBeValid'
         );
         // Adding the variant to $_SESSION to mark it as validated.
-        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)] = $_REQUEST['var'];
+        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)][$_REQUEST['var']] = true;
         exit;
     }
 
@@ -315,7 +315,7 @@ if ($_REQUEST['action'] == 'check') {
             'oButtonOKCouldBeValid'
         );
         // Adding the variant to $_SESSION to mark it as validated.
-        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)] = $_REQUEST['var'];
+        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)][$_REQUEST['var']] = true;
         exit;
     }
 
@@ -425,7 +425,7 @@ if ($_REQUEST['action'] == 'map') {
             'oButtonOKCouldBeValid'
         );
         // Adding the variant to $_SESSION to mark it as validated.
-        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)] = $_REQUEST['var'];
+        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)][$_REQUEST['var']] = true;
         exit;
     }
 
@@ -546,7 +546,7 @@ if ($_REQUEST['action'] == 'map') {
         }');
 
         // Adding this variant to the array of validated variants to mark it as validated.
-        $_SESSION['VV']['validated_variants'][$sTranscript] = $aTranscriptData['DNA'];
+        $_SESSION['VV']['validated_variants'][$sTranscript][$aTranscriptData['DNA']] = true;
     }
 
     // Returning the mapping for genomic variants.
@@ -591,7 +591,7 @@ if ($_REQUEST['action'] == 'map') {
         ');
 
         // Adding the variant to the array of validated variants to mark it as validated.
-        $_SESSION['VV']['validated_variants'][$sGBID] = $sMappedGenomicVariant;
+        $_SESSION['VV']['validated_variants'][$sGBID][$sMappedGenomicVariant] = true;
     }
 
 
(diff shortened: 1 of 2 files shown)


src/ajax/check_hgvs_dialogue.php
```diff
@@ -572,9 +589,13 @@ function update_images_per_step($sStep, $sImage)
         oGenomicField.val("' . $sMappedGenomicVariant . '").attr("class", "accept");
         oGenomicField.siblings("img:first").attr({src: "gfx/check.png", title: "Validated"});
         ');
+
+        // Adding the variant to the array of validated variants to mark it as validated.
+        $_SESSION['VV']['validated_variants'][$sGBID] = $sMappedGenomicVariant;
     }
 
 
+    // TODO: Check whether the below code does not conflict with the $_SESSION enforce method.
```
loeswerkman, line 567: The code below fills in fields that were left empty after a call to VariantValidator and disables there `onChange`. This was done because the original code worked in a way where users could not fill in the empty fields without there being a call to the `check_hgvs_dialogue.php` that would then empty all fields that were just nicely filled by VariantValidator. However, because our new code does not clear out all variants, the `onChanges` may remain. However, filling in the fields with _some_ sort of input seems clean to me, so I have left that. @ifokkema, if you disagree on filling the empty fields, that is totally fine and it can be removed!

ifokkema (changes requested):
Some changes are needed, I also found one XSS vulnerability that needs to be fixed.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -35,17 +35,30 @@
 header('Content-type: text/javascript; charset=UTF-8');
 
 
+// Check whether all required input was given.
+if (!(isset($_REQUEST['var']) && isset($_REQUEST['action'])
+    && isset($_REQUEST['fieldName']) && isset($_REQUEST['refSeqInfo']))) {
+    // If any of these variables are missing, we cannot correctly
+    //  perform any checks, so we will exit the script.
+    exit;
+}
```
ifokkema, line 44: This is fine, no need to change it, but just FYI: a perhaps more readable method is
```php
if (empty($_REQUEST['var']) || empty($_REQUEST['action'])
    || empty($_REQUEST['fieldName']) || empty($_REQUEST['refSeqInfo'])) {
```

src/ajax/check_hgvs_dialogue.php
```diff
@@ -35,17 +35,30 @@
...
+if (!isset($_REQUEST['transcripts'])) {
+    // Let's assume the transcripts were empty if this variable was not set.
+    $_REQUEST['transcripts'] = '';
+}
+
+
```
ifokkema, line 50: The line below already handles the missing variable, so this part is redundant.
(note that `empty($var)` is like an `!isset($var) || !$var`)

src/ajax/check_hgvs_dialogue.php
```diff
@@ -35,17 +35,30 @@
...
-$sFieldName = htmlspecialchars(urldecode($_REQUEST['fieldName']));
+$sFieldName = htmlspecialchars($_REQUEST['fieldName']);
 
 
 
-if (!($_REQUEST['var'])) {
+if (!empty($_REQUEST['var'])) {
     // If the variant is empty, we can simply close the script.
     exit;
 }
```
ifokkema, line 64: This kills the script entirely :laughing:

ifokkema: By the way, this block can now be removed because you already check for an empty variable on line 39.

src/ajax/check_hgvs_dialogue.php
```diff
@@ -41,7 +41,7 @@
 $aTranscripts = (empty($_REQUEST['transcripts'])? array() : explode('|', urldecode($_REQUEST['transcripts'])));
 
 // Retrieving the name of the input field.
-$sFieldName = urldecode($_REQUEST['fieldName']);
+$sFieldName = htmlspecialchars(urldecode($_REQUEST['fieldName']));
```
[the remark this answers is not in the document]
loeswerkman: Done

src/ajax/check_hgvs_dialogue.php
```diff
@@ -233,11 +209,13 @@ function update_images_per_step($sStep, $sImage)
...
         update_images_per_step('statusChecks', 'gfx/cross.png');
         update_dialogue(
-            '<br>Your variant contains syntax which our HGVS check cannot recognise. ' .
+            '<br>Your variant contains syntax that our HGVS check cannot recognise. ' .
             'Therefore, we cannot validate your variant nor map it to other reference sequences. ' .
             'Please thoroughly validate your variant by hand.',
             'oButtonOKCouldBeValid'
         );
+        // Adding the variant to $_SESSION to mark it as validated.
+        $_SESSION['VV']['validated_variants'][($sType == 'VOG'? $sGenomeBuildID : $sReferenceSequence)][$_REQUEST['var']] = true;
```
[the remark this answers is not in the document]
loeswerkman: Yes, good point! Has been implemented

src/class/object_genome_variants.php
```diff
@@ -256,21 +256,44 @@ function checkFields ($aData, $zData = false, $aOptions = array())
...
-                if (lovd_isHGVS($sVariant)) {
-                    // The variant looks good!
-                    continue;
+                // We want to check the input of all DNA fields: if their description passed
+                //  all our HGVS checks, it will have been added to $_SESSION.
+
+                // Before we can check $_SESSION, we need to retrieve the key to which it
+                //  should have been added, which is the GB ID for a VOG and the reference
+                //  sequence for a VOT.
+                global $_DB;
```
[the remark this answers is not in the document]
loeswerkman: Whoops, done!

src/ajax/check_hgvs_dialogue.php
loeswerkman: Much better, this has been implemented instead :)

src/ajax/check_hgvs_dialogue.php
loeswerkman: Whoops ':), the code block has been removed.

src/ajax/check_hgvs_dialogue.php
loeswerkman: Has been removed!

PR #613, opened 2022-09-13
Fix the HGVS validation interface

ifokkema:
Fix the HGVS validation interface.
- Fix that VV errors weren't recognized by the checkHGVS interface.
- Prevent notices when VV returns false.
- Mark NGs as unsupported since VV doesn't really support them now.
  - The VV library also doesn't support them, but when trying to build code that would handle them, I ran into so many issues that it's best if VV improves the NG support, and then I'll be able to work with them. LOVD doesn't use them, so it's OK.
  - See https://github.com/openvar/variantValidator/issues/423
- Mark LRGs (genome-based) as unsupported just like NG.
  - VV handles LRGs the same as NGs. There are currently too many issues to deal with, so communicate that they are currently unsupported.
- Improve handling of LRGt variants in the VV library.
  - It'll make sure that we select the correct output, and it'll suppress a few warnings that we don't want to see in these cases.
- Add a link to api.lovd.nl from the tabs in the menu.


src/class/variant_validator.php
```diff
@@ -917,7 +917,21 @@ public function verifyVariant ($sVariant, $aOptions = array())
...
-        $aData['data']['DNA'] = $aJSON['hgvs_transcript_variant'];
+        // Handle LRGt submissions.
+        if (substr($sVariant, 0, 3) == 'LRG') {
+            $aData['data']['DNA'] = $aJSON['hgvs_lrg_transcript_variant'];
+            // Also, in this case, we're not interested if new transcripts exist.
+            $aJSON['validation_warnings'] = array_filter(
+                $aJSON['validation_warnings'],
+                function ($sValue)
+                {
+                    return preg_match('/^Reference sequence .+ can be updated to .+\.$/', $sValue);
```
ifokkema, line 928: This return statement should be negated, no?

ifokkema (commented):
Better double-check this.

src/class/variant_validator.php
ifokkema: Yes; fixed it.

